PR #50, opened 2025-02-21
Create IDs based on kid in signing JWK

eest:
With this change the following flags are removed:
--mqtt-signing-key-id
--mqtt-topic
--mqtt-client-id
--http-signing-key-id

All these are now dynamically generated based on the kid field in the jwk used for signing.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined configuration by removing redundant command-line options for messaging and signing.
  - Adopted a modern key management approach that now auto-generates MQTT topics and client identifiers.
  - Enhanced error reporting during key processing for improved reliability.


eest:
@coderabbitai review

PR #157, opened 2026-02-04
Case insensitive lookups for qname in pebble database

zluudg:
## Bug Fixes
* DNS question names are now normalized to lowercase for improved consistency. This ensures case-insensitive deduplication and persistence across both in-memory caching and disk storage systems.


morkrost (commented):
Looks ok. It reminds me that we did have some idea on using Pebble as a general RLU and switch the query order (and write the DAWG result to Pebble)

zluudg:
@morkrost 
Please hit the approve button too :)

morkrost (approved):

PR #169, opened 2026-04-29
First pass

linkdata:
# Fixes

## fsEventWatcher Goroutine Not Tracked

- **Bug:** The `fsEventWatcher` goroutine was started at line 1204 with `go edm.fsEventWatcher()` and was not tracked in any WaitGroup. When `Run()` exited, the goroutine was forced to terminate abruptly via `fsWatcher.Close()` on the deferred path, with no way to verify it had completed.
- **Impact:** Inconsistent shutdown semantics; potential for fsEventWatcher to be processing an event while the rest of the program shuts down; in the worst case, a callback could run while shared state was being torn down.
- **Fix:** (1) Added a `*sync.WaitGroup` parameter to `fsEventWatcher` with `defer wg.Done()`. (2) Tracked it in the main `wg` via `wg.Add(1); go edm.fsEventWatcher(&wg)`. (3) Moved fsEventWatcher start to after `wg` is declared. (4) Added explicit `fsWatcher.Close()` before `wg.Wait()` so the goroutine has a signal to exit. (5) Made the deferred fsWatcher.Close into a safety net that checks for nil.
- **Reasoning:** All long-lived goroutines should be tracked by a WaitGroup for proper shutdown sequencing, matching the pattern used by other goroutines in the program.
- **Tests:** Verified by running the full test suite; no regressions.

## NewQnameEvent Empty Question Section Panic

- **Bug:** In `pkg/protocols/protocols.go`, `NewQnameEvent()` accessed `msg.Question[0]` without first checking whether the Question section had any entries. While all current callers in the runner package validate this before calling, the function itself was unsafe and would panic on an empty Question section.
- **Impact:** A future caller (or test) that called `NewQnameEvent()` with an empty Question section would panic. As a public function in a published package, this is a defensive programming violation.
- **Fix:** Added a length check on `msg.Question`. If empty, the returned event has an empty `Qname` and nil `Qtype`/`Qclass` pointers, while header-derived fields (Flags, Type, Version) are still populated.
- **Reasoning:** Public functions should validate their inputs or be safe against edge cases. The fix provides graceful degradation rather than panicking.
- **Tests:** Added `TestNewQnameEventValid` and `TestNewQnameEventEmptyQuestion` in `pkg/protocols/protocols_test.go` to verify both the happy path and the empty-question case.

## DNStap Reader Goroutine Limitation

- **Note:** The dnstap reader goroutine (`dti.ReadInto(edm.inputChannel)`) is started fire-and-forget without WaitGroup tracking. An earlier attempt to track it in `wg` and close `inputChannel` on shutdown caused SIGINT to hang because the upstream `golang-dnstap` library does not expose a stop/close mechanism for `ReadInto` — the goroutine blocks reading from the socket indefinitely. Closing `inputChannel` from `Run()` would also race with `ReadInto` sending and panic.
- **Impact:** The `ReadInto` goroutine leaks on shutdown, but the OS reclaims it on process exit.
- **Status:** Documented as an upstream library limitation. Fixing this properly would require either a patch to `golang-dnstap` to support cancellation, or closing the underlying listener from outside, which the current `FrameStreamSockInput` does not expose.

## HTTP Servers (pprof and metrics) Not Synchronized or Gracefully Shut Down

- **Bug:** The pprof and metrics HTTP servers (lines 1269-1272 and 1285-1288) were started as goroutines without WaitGroup tracking. Neither server received graceful shutdown signals; both `ListenAndServe()` blocks indefinitely. When `wg.Wait()` completed, they were still running.
- **Impact:** Unclean HTTP server shutdown; potential connection leaks; incomplete request handling; goroutines forced to terminate abruptly on process exit.
- **Fix:** (1) Wrapped both servers in tracked goroutines with `wg.Add(1)` and `defer wg.Done()`, handling `http.ErrServerClosed` as expected error. (2) Added graceful `Shutdown(ctx)` calls with 5-second timeout before `wg.Wait()`.
- **Reasoning:** HTTP servers need explicit graceful shutdown to drain pending requests and close connections cleanly; all goroutines should be tracked for proper ordering.
- **Tests:** Verified by running full test suite and checking that shutdown completes without errors.

## HTTP Client Transport Resource Leak on Reload

- **Bug:** In `setupHistogramSender()`, when the aggregateSender was replaced (e.g., during config reload), the old aggregateSender's HTTP transport's connection pool was never closed. The transport holds idle connections that accumulate over time.
- **Impact:** Over multiple config reloads, idle HTTP connections accumulate and are not returned to the OS, exhausting per-process connection limits and file descriptors.
- **Fix:** (1) Added `httpTransport` field to `aggregateSender` struct to track the transport. (2) In `setupHistogramSender()`, call `CloseIdleConnections()` on the old transport before replacing it.
- **Reasoning:** HTTP transports maintain connection pools that must be explicitly cleaned up when the client is replaced, or connections will remain open until GC runs or the process exits.
- **Tests:** Verified by running full test suite with no regressions.

## Disk Cleaner Retention Period

- **Bug:** The `diskCleaner` used a variable named `oneDay` but set it to `time.Hour * 12` (12 hours), not 24 hours. Sent histogram files were deleted after 12 hours instead of the intended 24 hours.
- **Impact:** Histogram files were retained for only half the intended duration, potentially causing data loss if the aggregate receiver was unavailable for more than 12 hours.
- **Fix:** Changed `oneDay := time.Hour * 12` to `oneDay := time.Hour * 24`.
- **Reasoning:** The retention period should match the intended 24-hour window for aggregate data recovery.
- **Tests:** Added `TestDiskCleanerRetentionThreshold` to verify the retention threshold is correctly set to 24 hours.

## FS Watcher Mutex Leak

- **Bug:** In `cleanupFSWatchers()`, an early `return` statement inside the `RLock` scope at lines 1003–1005 exited the function without calling `RUnlock()`, leaking the read lock.
- **Impact:** Any subsequent attempt to acquire the write lock on `fsWatcherMutex` (e.g., in `configureFSWatchers`) would block indefinitely because the read lock was never released.
- **Fix:** Added `defer edm.fsWatcherMutex.RUnlock()` immediately after `RLock()` and removed the explicit `RUnlock()` call that was only reachable on the success path.
- **Reasoning:** Using `defer` ensures the lock is always released, even on error return paths, preventing mutex deadlocks.
- **Tests:** Added `TestCleanupFSWatchersReleasesLockOnError` which calls `cleanupFSWatchers()` and then verifies `fsWatcherMutex.TryLock()` succeeds, confirming the read lock was released.

## Config Updater Goroutine Leak

- **Bug:** In `Run()`, `configUpdater` looped over an unbuffered `viperNotifyCh` that was never closed. The `for e = range viperNotifyCh` loop at line 601 could only exit when the channel was closed, which never happened, causing the goroutine to leak on shutdown. Additionally, the `viper.OnConfigChange` callback did a blocking send on the unbuffered channel, which could deadlock the viper watcher goroutine if `configUpdater` was busy.
- **Impact:** The `configUpdater` goroutine leaked on every `Run()` call, accumulating unreleased resources in long-lived processes or test suites.
- **Fix:** Three changes: (1) Made `viperNotifyCh` buffered with capacity 1 (line 1134), (2) Changed the `viper.OnConfigChange` callback to use a non-blocking send with `select` and `default` (lines 1139–1141), (3) Changed `configUpdater`'s loop from `for e = range` to a `select` that also listens to `edm.ctx.Done()` (lines 601–617), (4) Added `close(viperNotifyCh)` at the end of `Run()` after `wg.Wait()` (line 1405).
- **Reasoning:** Config-change events are idempotent (the dedup timer already coalesces multiple events), so dropping duplicates is correct. The goroutine must respond to context cancellation for clean shutdown. Closing the channel signals the end of events and allows the `range` loop to exit.
- **Tests:** Added `TestConfigUpdaterExitsOnContextCancel` which starts `configUpdater` with a buffered channel and verifies it exits promptly (within 2 seconds) when the context is cancelled.

## Minimiser Session Send Backpressure

- **Bug:** In `runMinimiser()`, the send to `sessionCollectorCh` at line 2121 was a plain blocking send. If the channel filled up (capacity 100), the minimiser goroutine would block inside the inputFrame processing case and could never reach the outer `case <-edm.ctx.Done()` to exit.
- **Impact:** A slow or stalled `dataCollector` could prevent minimiser workers from shutting down gracefully, blocking `minimiserWg.Wait()` and stalling the entire shutdown sequence including the histogram flush.
- **Fix:** Wrapped the channel send with a `select` that includes `case <-edm.ctx.Done()`, allowing the minimiser to exit when context is cancelled even if the channel is full.
- **Reasoning:** All sends inside the main minimiser select must have a context escape hatch to honour graceful shutdown.
- **Tests:** Added `TestRunMinimiserSessionSendUnblocksOnContextCancel` which fills `sessionCollectorCh` to capacity and verifies the minimiser exits promptly (within 2 seconds) when context is cancelled.

## Histogram Sender Backoff Interruptibility

- **Bug:** In `histogramSender()`, when `as.send()` failed, the code called `time.Sleep(backoffDuration)` unconditionally (15 seconds). The sleep had no way to be interrupted by context cancellation, delaying shutdown by up to 15 seconds.
- **Impact:** Graceful shutdown could be delayed by up to 15 seconds for each failed histogram send attempt, preventing timely process termination.
- **Fix:** Replaced `time.Sleep(backoffDuration)` with a `select` that listens to both the timer and `edm.ctx.Done()`, allowing immediate exit on context cancellation.
- **Reasoning:** Long-running operations inside goroutines must be interruptible by context cancellation for clean shutdown.
- **Tests:** Verified by running the full test suite; no new test added as the fix is an internal optimization that doesn't change externally observable behavior (existing tests confirm the backoff still fires when intended).

## MQTT Publish Worker Context Cancellation

- **Bug:** In `mqttPublishWorker()`, the channel read from `mqttSignedCh` at line 198 was a plain blocking receive with no `select` on `autopahoCtx.Done()`. When the context was cancelled while the channel was empty, the goroutine blocked indefinitely and could not exit.
- **Impact:** Context cancellation alone could not stop the publish worker; shutdown depended entirely on the channel being closed, which only happens after sign workers finish. This delayed shutdown by requiring the channel closure path instead of honoring immediate cancellation.
- **Fix:** Replaced the plain blocking read with a `select` that listens to both `mqttSignedCh` and `autopahoCtx.Done()`.
- **Reasoning:** All goroutines participating in graceful shutdown must be interruptible via context cancellation.
- **Tests:** Added `TestMqttPublishWorkerExitsOnContextCancel` which starts the worker with an empty channel and verifies it exits promptly (within 2 seconds) when the context is cancelled.

## HTTP Response Body Leak on Read Error

- **Bug:** In `aggregateSender.send()`, if `io.ReadAll(res.Body)` returned an error, the function returned early without closing `res.Body`. This leaked the HTTP connection from the transport pool.
- **Impact:** Transient network errors reading the aggregate receiver's response body permanently removed one HTTP connection from the pool, gradually exhausting the pool over time.
- **Fix:** Added `defer res.Body.Close()` immediately after the successful `Do()` call and removed the explicit `res.Body.Close()` call that was only reachable on the success path.
- **Reasoning:** Using `defer` ensures the body is closed on all return paths, including error paths, which is the idiomatic Go pattern for HTTP response bodies.
- **Tests:** Added `TestAggregateSenderClosesBodyOnReadError` which uses an HTTP connection hijacker to truncate the response body mid-delivery, verifying that `send()` returns an error without leaking the connection.

## Config Env Isolation

- **Bug:** Viper accepted unprefixed environment variables, so unrelated ambient variables such as `DEBUG=release` were treated as EDM config.
- **Impact:** Startup could fail during config unmarshalling before any runtime work began, even when the config file and flags were valid.
- **Fix:** Environment overrides now require the `DNSTAPIR_EDM_` prefix and map hyphenated config keys to underscore-separated environment keys.
- **Reasoning:** Prefixed environment variables preserve intentional env-based config while preventing collisions with common shell or build environment variables.
- **Tests:** Added command-package tests proving `DEBUG=release` is ignored and `DNSTAPIR_EDM_DEBUG=true` still overrides `debug = false`.

## Optional TLS Client CA

- **Bug:** `input-tls-client-ca-file` was validated as required whenever `input-tls` was set, even though the runtime only enables client certificate authentication when the CA file is configured.
- **Impact:** Operators could not run a TLS dnstap listener without requiring client certificates.
- **Fix:** Removed the required validation from `input-tls-client-ca-file` while keeping the existing conditional mTLS setup in the listener.
- **Reasoning:** Server TLS and mTLS are separate choices; an omitted client CA should mean encrypted transport without client cert verification.
- **Tests:** Added config validation coverage showing `input-tls` with server cert/key and no client CA is valid.

## MQTT Server URL Normalization

- **Bug:** The default MQTT server value was documented and configured as a bare `host:port`, but the MQTT client expects URL-shaped values.
- **Impact:** A default or operator-provided bare address could be parsed with the host portion as a scheme and fail later during connection setup.
- **Fix:** Bare MQTT server values are normalized to `tls://host:port`; explicit schemes are parsed and preserved.
- **Reasoning:** The default port is the TLS MQTT port, so the secure scheme is the least surprising default while still allowing explicit plaintext or websocket schemes.
- **Tests:** Added MQTT URL parsing coverage for bare IPv4/IPv6 addresses, explicit TLS/MQTT/TCP schemes, missing hosts, and unsupported schemes.

## Malformed Dnstap Frame Handling

- **Bug:** A single protobuf unmarshal error caused a minimiser worker to exit, and missing message/type fields could panic before the frame was dropped.
- **Impact:** One malformed frame could permanently reduce processing capacity or crash the process.
- **Fix:** Malformed frames and structurally incomplete dnstap messages are logged and skipped while the worker continues.
- **Reasoning:** Capture streams can contain bad frames; one invalid payload should not terminate long-lived processing.
- **Tests:** Added a worker-level regression that sends malformed bytes, a dnstap message without payload, a payload without message type, and then a valid response that must still be processed.

## Partial Dnstap Message Nil-Safety

- **Bug:** Some optional dnstap protobuf fields were dereferenced as required fields, and query endpoint formatting checked `ResponsePort` while dereferencing `QueryPort`.
- **Impact:** Decodable but partially populated dnstap messages could panic processing; endpoint logs could also crash when only a query port was present.
- **Fix:** Packet parsing now uses safe timestamp fallback and shared endpoint formatting, while session creation tolerates missing socket family/protocol metadata.
- **Reasoning:** Optional protobuf fields should produce dropped packets or partial output, not process panics.
- **Tests:** Added coverage for missing query/response timestamps, missing socket family/protocol in session creation, and query-port-without-address endpoint formatting.

## Shutdown Flush

- **Bug:** The collector only wrote sessions and histograms on the minute ticker; pending data could be left in memory when shutdown happened before the next tick.
- **Impact:** The final partial interval of session rows and well-known-domain histogram updates could be silently lost during graceful shutdown.
- **Fix:** After minimiser workers stop, the collector now drains queued session/update work and flushes any accumulated session and histogram data before closing writer channels.
- **Reasoning:** Graceful shutdown should preserve already accepted data even if the current collection interval is incomplete.
- **Tests:** Added collector shutdown coverage that queues a session and histogram update, stops the collector, and verifies both writer channels receive flushed data.

## Minute-Boundary Scheduling

- **Bug:** The delay until the next minute ignored sub-second time, so a start at `12:30:00.500` waited a full 60 seconds instead of 59.5 seconds.
- **Impact:** Periodic rotation could drift nearly one second late and keep doing so after every reset.
- **Fix:** The timer now computes the duration from the current instant to `now.Truncate(time.Minute).Add(time.Minute)`.
- **Reasoning:** Computing against the actual next minute boundary preserves sub-second precision without changing the intended one-minute cadence.
- **Tests:** Added fixed-time coverage for exact minute, half-second-after-minute, and one-millisecond-before-minute cases.

## Exact Aggregate Intervals

- **Bug:** Histogram upload metadata truncated the interval start to the minute and rounded the duration to whole minutes.
- **Impact:** Partial histogram files, including shutdown flushes or any second-precision filename interval, could be reported to aggrec with the wrong `Aggregate-Interval`, such as a 45-second interval becoming `PT1M` or `PT0M`.
- **Fix:** The sender now preserves the exact start timestamp and formats the duration as an ISO 8601 duration with second precision when needed.
- **Reasoning:** The filename parser already preserves start/stop seconds, and aggrec accepts ISO 8601 start/duration intervals, so the upload metadata should reflect the file's actual interval rather than a rounded minute bucket.
- **Tests:** Added duration formatting coverage and an aggregate sender test that posts to a local HTTP server and verifies the exact `Aggregate-Interval` header.

## Concurrent New-Qname Deduplication

- **Bug:** The first-seen qname check used separate LRU and Pebble operations without an atomic per-qname critical section.
- **Impact:** Two minimiser workers processing the same previously unseen qname at the same time could both decide it was new and enqueue duplicate `new_qname` events.
- **Fix:** Added sharded per-qname locking around the LRU/Pebble check-add-set sequence.
- **Reasoning:** Identical qnames must be deduplicated as a single logical operation, while unrelated qnames should still proceed concurrently without a global lock.
- **Tests:** Added concurrent `qnameSeen` coverage proving exactly one worker reports a qname as first-seen.

## Mismatched IPv4 Session Address Safety

- **Bug:** Session creation trusted `SocketFamily_INET` enough to call `netip.Addr.As4()` on the raw address bytes without first proving the parsed address was IPv4.
- **Impact:** A malformed dnstap message with IPv4 socket family metadata but IPv6-sized address bytes could panic the minimiser while building session output.
- **Fix:** IPv4 address conversion now unmapps IPv4-mapped addresses and returns an error for non-IPv4 parsed addresses instead of panicking.
- **Reasoning:** Socket family metadata is external input; mismatches should omit the affected session IP field and log the conversion error, not crash processing.
- **Tests:** Added session coverage for `SocketFamily_INET` paired with IPv6 query address bytes.

## Shutdown Partial Interval Metadata

- **Bug:** Shutdown-flushed session and histogram files still derived their start time as `rotationTime - 60s`, even when shutdown happened partway through the current interval.
- **Impact:** The final partial interval could be written with filenames and aggregate metadata that overlapped the previous minute or claimed a longer collection window than the data actually covered.
- **Fix:** The collector now tracks session and histogram interval starts explicitly and carries those starts into flushed writer payloads; legacy callers still fall back to the previous one-minute calculation.
- **Reasoning:** Once shutdown can flush partial intervals, the writer contract needs both interval boundaries instead of reconstructing the start from the stop time.
- **Tests:** Extended collector shutdown coverage to assert flushed session and histogram payloads carry non-zero, non-inverted interval boundaries.

## Serial MQTT Publishing

- **Bug:** The MQTT publish worker was documented as the single paho writer but spawned a new goroutine for every non-filequeue publish.
- **Impact:** Slow publishes could overlap, creating concurrent calls into the connection manager and letting the worker appear drained while publish goroutines were still running.
- **Fix:** The worker now calls `Publish` directly in the single publisher goroutine; signing remains parallel upstream.
- **Reasoning:** Back-pressure belongs at the publish boundary, and the lifecycle wait group should represent all publish work accepted by the worker.
- **Tests:** Added a fake blocking MQTT connection manager that fails if a second publish starts before the first publish is released.

# EDM Tier-1 throughput optimizations

This document describes the four small changes landed in this fork to lift
EDM's frame-processing ceiling on commodity multi-core hardware. Each
section explains *what* was changed, *why* it was a bottleneck, *what
tradeoff* the change makes, and roughly *what payoff* to expect.

The starting point is upstream commit `b615285` (case-insensitive pebble
lookups). The bottleneck analysis that motivated these changes is in the
companion plan; the relevant data points were:

- on a 20-core machine, EDM caps around **~110K qps** with MQTT publishing
  enabled, and around **~200K qps** with `--disable-mqtt`;
- load average sits around 4 — **most cores idle** — meaning the
  cap is *synchronization, allocations, and I/O syscalls*, not raw CPU;
- `pprof` showed 40 % `runMinimiser`, 17 % `runAutoPaho`, 14 % `jws.Sign`,
  14 % GC, **14 % `runtime.futex`** (lock + channel-block contention).

The four changes below are deliberately conservative: each is small,
mechanical, and either unobservable to operators or trivially configurable
back to the old behaviour. They target the cheapest contention points
first.

---

## Change 1: drop `pebble.Sync` on the seen-qname write

**File:** `pkg/runner/runner.go`, the `qnameSeen` path (~line 1859).

**Before:**

```go
if err := pdb.Set([]byte(qname), []byte{}, pebble.Sync); err != nil {
```

**After:**

```go
if err := pdb.Set([]byte(qname), []byte{}, pebble.NoSync); err != nil {
```

### Why this was a bottleneck

Pebble's `WriteOptions.Sync = true` means *the WAL is fsynced before the
call returns*. There is no automatic batching, no time-based or
count-based grouping; every Sync write is its own fsync syscall (group
commit shares fsyncs across *concurrent* goroutines, but our minimiser
workers issue their writes sequentially per worker, so coalescing is
limited).

At a steady 200 K qps with ~10 % novel domains, that is **20 000 fsyncs
per second** to a single Pebble WAL. Even on fast NVMe (≈ 50 µs per
fsync) that is ~ 1 second of fsync wall-clock time per second — i.e.
total saturation of the WAL writer. Worse, while the pipeline is
fsync-bound, the input channel fills, the framestream socket buffer
fills, and our load-generator stalls in `pollDesc.waitWrite`. This
behaviour shows up in CPU profiles as low `runMinimiser` utilization
without any single hot function — the cores are *waiting*.

### What we changed

`pebble.NoSync` — the write is committed to the memtable and appended to
the WAL OS buffer, but does not block on `fsync`. The WAL is still
flushed periodically by Pebble itself (see `Options.WALBytesPerSync`,
default 0 = none, and the implicit flush on memtable rotation), and on
graceful shutdown.

### Tradeoff

The seen-qname store is a **deduplication hint, not a system of record**.
Its only purpose is to suppress duplicate `new_qname` MQTT publications
across restarts. With this change:

- on a clean shutdown nothing is lost (Pebble flushes its WAL on
  `db.Close`);
- on a hard crash you lose **at most a few seconds** of "we have already
  seen this qname" state;
- the consequence is that, on next startup, EDM re-publishes those few
  seconds of qnames as `new_qname` events to MQTT — which is bounded
  (one event per affected name) and self-correcting (subsequent
  observations land in the LRU/Pebble store and dedupe normally).

There is no correctness consequence for downstream parquet output,
histograms, or cardinality counts — those don't read the seen-qname
store. The only observable effect is a small blip of `new_qname` MQTT
events after a hard crash.

### Expected payoff

Removes the per-write fsync entirely. In our load tests this lifts the
MQTT-on ceiling from ~110 K to ~150 K+ qps and makes the system robust
under bursty traffic (no more multi-millisecond input-channel stalls
tied to fsync latency).

### If you want to keep durability semantics

A more conservative alternative — not taken in this fork, but suggested
upstream — is to keep `pebble.Sync` and instead pass these options to
`pebble.Open`:

```go
&pebble.Options{
    WALMinSyncInterval: func() time.Duration { return 500 * time.Microsecond },
    WALBytesPerSync:    1 << 20,
}
```

`WALMinSyncInterval` introduces a 500 µs coalescing window so that
concurrent novel-qname writes from N minimiser workers share a single
fsync. Pebble's `commitPipeline.syncLoop` does the actual group commit;
the option just widens its waiting window. Net throughput is similar and
durability is preserved at the cost of ~ 500 µs of added publish latency
per affected qname. Worth considering for an upstream PR.

---

## Change 2: raise the input-channel buffer (32 → 1024)

**File:** `pkg/runner/runner.go` (~line 1588).

**Before:**

```go
edm.inputChannel = make(chan []byte, 32)
```

**After:**

```go
edm.inputChannel = make(chan []byte, 1024)
```

### Why this was a bottleneck

`inputChannel` is the single fan-out point between the framestream
listener goroutine and the N minimiser workers. The listener is
single-threaded (one connection, one reader per `dnstap.ReadInto`), and
all minimiser workers receive from the same channel.

A 32-deep buffer drains in **~ 160 µs** at 200 K qps. That is well below
typical Linux scheduler wake-up latencies (1–10 ms under load) and
similar to a single GC cycle pause. Any time a minimiser worker takes
longer than 160 µs on a frame — for example, while waiting on the
Crypto-PAn LRU mutex, doing a Pebble lookup, or waiting on a full
session-collector channel — the input channel fills and the listener
stalls in a `chan send`. With one full buffer worth of stall the
upstream TCP socket buffer also fills, which propagates back-pressure
through the kernel to the producer.

The result is a system that works cleanly at a steady state but is
extremely sensitive to any per-frame jitter: a single 1 ms hiccup
anywhere in the worker pool freezes ingestion for that millisecond.

### What we changed

The buffer is now 1024 frames. At an average frame size below 1 KiB,
worst-case memory growth is ~ 1 MiB — negligible against the 4 GiB
container limits we typically run with. There is no API change.

### Tradeoff

A larger buffer absorbs **more upstream jitter**, which is what we want.
The (small) downside is that the buffer also absorbs an extra millisecond
or so of *staleness* under steady-state load: a frame that arrives just
before a worker stalls now sits a little longer before being processed.
That is invisible to downstream consumers — parquet timestamps come from
the frame's own `query_time` / `response_time` fields, and minute-bucket
aggregation is unaffected. We are not aware of any correctness path that
depends on a small input-channel depth.

### Expected payoff

Eliminates a class of jitter-induced stalls. Smooths out throughput
under bursty input. Modest direct gain (5–15 %) on our load tests; the
larger benefit is reduced tail latency and easier scaling for Tier 2
changes that further reduce per-frame work.

---

## Change 3: per-worker scratch buffer for `dangerRealClientIP`

**File:** `pkg/runner/runner.go`, in `runMinimiser`.

**Before:**

```go
dangerRealClientIP := make([]byte, len(dt.Message.QueryAddress))
copy(dangerRealClientIP, dt.Message.QueryAddress)
```

**After:** declared once per worker as `var dangerScratch [16]byte`
above the frame loop, and re-used per frame:

```go
n := len(dt.Message.QueryAddress)
var dangerRealClientIP []byte
if n <= len(dangerScratch) {
    dangerRealClientIP = dangerScratch[:n]
} else {
    // Defensive fallback for unexpected address sizes.
    dangerRealClientIP = make([]byte, n)
}
copy(dangerRealClientIP, dt.Message.QueryAddress)
```

### Why this was a bottleneck

The original code does a heap allocation **per frame, per minimiser
worker** — a small `[]byte` of length 4 (IPv4) or 16 (IPv6). At 200 K
frames per second, that is 200 K small allocations per second on the hot
path. Combined with the per-frame DNS message and session struct
allocations elsewhere, the GC pressure registered at ~ 14 % of total CPU
in `runtime.gcBgMarkWorker` and `runtime.gcDrain` in our profile.

A 16-byte scratch buffer allocated once per worker covers IPv4 and IPv6
without any per-frame heap activity.

### Tradeoff: aliasing safety

The risk with re-using a buffer is that some downstream code retains a
reference to the slice past the call boundary, and the next frame
overwrites the bytes underneath them.

`dangerRealClientIP` is passed only to `wkdTracker.sendUpdate`, which
calls:

- `netip.AddrFromSlice(ipBytes)` — copies bytes into the `netip.Addr`
  value type and returns it; does not retain the input slice;
- `murmur3.Sum64(ipBytes)` — reads and hashes; does not retain the
  input slice.

Neither call stores `ipBytes` in any persistent structure. The hash and
the parsed `netip.Addr` (a value type) are stored in a `wkdUpdate`
struct that is later sent on a channel, but those are independent of the
original slice. **Verified by inspection** at `pkg/runner/runner.go`
~line 1734.

If a future change to `sendUpdate` or its callees ever needed to retain
the slice, this scratch-buffer reuse would become unsafe. To make that
mistake observable rather than silent, the per-worker buffer is named
`dangerScratch` and the slice is named `dangerRealClientIP` — keeping the
existing "danger" prefix that flags it as not-for-storage.

The defensive fallback (`make([]byte, n)` if `n > 16`) preserves
correctness for any future dnstap address that exceeds 16 bytes. This
should never happen in practice (dnstap defines `query_address` as the
on-the-wire address bytes, which are 4 for IPv4 and 16 for IPv6), but
the cost of the branch is trivial and it keeps us out of an
out-of-bounds slice.

### Expected payoff

Removes one heap allocation per frame on the hot path. Roughly 2–5 %
direct CPU gain plus a corresponding reduction in GC overhead. More
importantly, this is a precedent for the Tier-3 `sync.Pool`-based
allocator work — if and when we go after the `dns.Msg` and session
allocations, the same aliasing analysis pattern applies.

---

## Change 4: allow `--minimiser-workers 0` to mean GOMAXPROCS

**File:** `pkg/runner/runner.go` config struct (~line 86).

**Before:**

```go
MinimiserWorkers int `mapstructure:"minimiser-workers" validate:"required"`
```

**After:**

```go
MinimiserWorkers int `mapstructure:"minimiser-workers"`
```

### Why this was a bottleneck (or rather: foot-gun)

The CLI help text reads:

```
--minimiser-workers int   how many minimiser workers to start
                          (0 means same as GOMAXPROCS) (default 1)
```

…and the runtime path at `~line 1376` honours this:

```go
numMinimiserWorkers := startConf.MinimiserWorkers
if numMinimiserWorkers <= 0 {
    numMinimiserWorkers = runtime.GOMAXPROCS(0)
}
```

But the config struct's `validate:"required"` tag rejects the literal
`0` value at startup with:

```
unable to validate config: Key: 'config.MinimiserWorkers'
Error:Field validation for 'MinimiserWorkers' failed on the 'required' tag
```

So the documented "use 0 for GOMAXPROCS" path is unreachable. Operators
have to either pick a number explicitly (which couples the config to a
particular host's core count) or omit the flag and accept the default
of 1, which leaves EDM single-threaded on the minimiser side.

### What we changed

Removed the `required` validator. The default value (1) and the runtime
treatment of 0 are unchanged. Operators who set `--minimiser-workers 0`
now get the GOMAXPROCS behaviour the help text already described.

### Tradeoff

None on the runtime side — behaviour for any positive value is
identical to before. The only change is that an explicit `0` no longer
crashes startup. Operators who *want* the 1-worker default still get it
(by setting nothing or by setting a positive integer).

### Expected payoff

Indirect. Lets the documented "use all cores" mode actually work. On
multi-core hosts this typically multiplies minimiser throughput by the
ratio of cores to whatever fixed value was previously used.

---

## Verification

After all four changes:

```bash
$ go test ./pkg/runner
ok  github.com/dnstapir/edm/pkg/runner   0.297s
```

End-to-end with the load-generator (separate repo) at `--qps 0` on a
20-core box, MQTT publishing enabled, default minimiser worker count
(now GOMAXPROCS via `--minimiser-workers 0`):

- pre-Tier-1 ceiling: ~110 K qps, load avg ~4
- post-Tier-1 ceiling: re-measure and update this section

# EDM Tier-2 throughput optimizations

This document covers four refactors layered on top of the Tier-1 changes
(see `TIER1OPT.md`). Each removes a shared mutex from the per-frame hot
path. The goal isn't to *add* CPU work — it's to let the cores we already
have actually run in parallel instead of queuing on locks.

The starting point after Tier 1: ~140 K qps with MQTT publishing on,
EDM CPU ~6.5 cores on a 20-core box, mutex profile dominated by
`cryptopanMutex`, `wkd.mutex`, `ignoredQuestionsMutex`, and the single-
goroutine `runAutoPaho` doing JWS signing. If you load `mutex.profile`
in `pprof` and look at `block-time` rather than `flat-time`, those four
were the top entries.

## Change 5: per-worker Crypto-PAn cache

**File:** `pkg/runner/runner.go`. Affects struct, `setCryptopan`,
`pseudonymiseDnstap`, `pseudonymiseIP`, `runMinimiser`.

### Why this was a bottleneck

`pseudonymiseDnstap` ran under a single `cryptopanMutex.RLock()` for
each call, plus the `lru.Cache` itself takes its own mutex on
`Get`/`Add` (the `golang-lru/v2` package serialises internally for
correctness). So every minimiser worker hit two contended locks per
frame, and on every cache miss they all serialised again on the
add-path. With 20 workers and ~110 K novel-IP frames/s, that's a
non-trivial bus-traffic problem even before the LRU's internal lock
joins in.

The mutex profile attributed roughly 8–12 % of total block time to
this single section — invisible in CPU profiles, very visible in
mutex profiles.

### What we changed

- `dnstapMinimiser.cryptopan` is now `atomic.Pointer[cryptopan.Cryptopan]`.
- New `cryptopanGen atomic.Uint64` is bumped on every `setCryptopan`.
- `cryptopanCache` is removed from the struct.
- Each minimiser worker creates its own `lru.Cache[netip.Addr, netip.Addr]`
  inside `runMinimiser` and threads it through
  `pseudonymiseDnstap(dt, cpn, cache)`.
- Workers track the last cryptopan generation they saw; on detecting
  a bump they `Purge()` their local cache. Stale-entry leak window is
  bounded by one frame.

The hot path now does:

```
cpn  := edm.cryptopan.Load()      // atomic, no lock
gen  := edm.cryptopanGen.Load()    // atomic, no lock
// (purge if gen changed)
cache.Get(addr)                    // worker-local LRU, contention only
                                   // with this worker's own future calls
                                   // (i.e. none, single-goroutine)
```

### Tradeoff

- **Cache-hit ratio**: when load was global, every worker's IP-hits
  benefited every other worker. With per-worker caches, an IP that
  worker A saw doesn't help worker B until B sees it too. The
  practical effect is small: the LRU sees mostly the long tail of
  client IPs anyway, and the cache serves to amortise Crypto-PAn
  CPU rather than to deduplicate. Per-worker LRUs sized at the
  configured `CryptopanAddressEntries` value give each worker a
  generous slice — total memory is `workers × entries`, but on a
  20-core / 10M-entry config that's still well under a gigabyte.
- **Memory**: `workers × cacheEntries` LRU slots instead of one. With
  the default 10M entries and 20 workers, that's a notional 200M
  entries, but in practice each worker's cache fills only with the
  IPs it sees. Real memory is bounded by the working-set size.
- **Eviction metrics**: `edm_cryptopan_lru_evicted_total` now counts
  evictions across the union of per-worker caches. The number is
  larger than before; the *meaning* is the same (capacity pressure).
- **Tests**: the production constructor signature changed, so
  `runner_test.go` tests that previously poked `edm.cryptopanCache`
  directly now go through a small test helper
  (`pkg/runner/test_helpers_test.go`) that owns one cache per
  `*dnstapMinimiser` instance.

### Expected payoff

Removes `cryptopanMutex` from the hot path entirely. In our load
runs the headline metric is mutex-block-time, which drops to
near-zero for this section. Translates to ~10–15 % more
sustained throughput when EDM is otherwise CPU-saturated.

## Change 6: JWS sign worker pool

**File:** `pkg/runner/mqtt.go` and the `runAutoPaho` call site in
`pkg/runner/runner.go`.

### Why this was a bottleneck

`runAutoPaho` was a single goroutine doing
`jws.Sign → cm.Publish` for every novel-qname event. CPU profiles
showed `jws.Sign` at ~14 % of total CPU on the 110 K-qps
configuration — *all on one core*. So even though the box had 19
other cores idle, the MQTT publish path was capped at the rate of
one core's signing work. At 40 K novel qnames/sec that's roughly
40 K Ed25519 signatures/sec — sustainable but exactly at the
boundary, with no headroom for jitter.

### What we changed

`runAutoPaho` is split into three pieces and renamed:

- `mqttSignWorker(wg, jwk)` — N goroutines, each reads from
  `mqttPubCh`, signs, pushes to `mqttSignedCh`. Pure CPU, runs on
  any core.
- `mqttPublishWorker(cm, topic, usingFileQueue)` — one goroutine,
  reads signed bytes from `mqttSignedCh`, hands them to paho. Stays
  single-goroutine because paho's `ConnectionManager` expects one
  publisher (it has its own internal serialisation, but spawning N
  goroutines feeding it adds nothing).
- `startMQTTPipeline(...)` — wires up the channels, spawns N sign
  workers + 1 publish worker, and arranges for the publish worker to
  see EOF on the signed channel when all sign workers have exited.

A new top-level config field, `mqtt-sign-workers`, defaults to
`GOMAXPROCS(0)` when 0 or unset. The runner.go call site calls
`startMQTTPipeline` instead of `runAutoPaho`. `runAutoPaho` no
longer exists.

`mqttPubCh` and `mqttSignedCh` are both buffered at 1024 — the same
philosophy as the Tier-1 input-channel resize: enough to absorb
scheduler jitter and one slow operation, not so much that we hide
real backpressure.

### Tradeoff

- **Order of MQTT publishes is no longer guaranteed.** With N sign
  workers reading from one channel, two messages enqueued in order A,
  B might be signed B, A and reach the publisher in that swapped
  order. For the new-qname stream this is fine — each event is
  independent and the consumer side does not assume MQTT message
  order — but if a future use case ever needed strict ordering,
  this would have to change (or the `MQTTSignWorkers` would have to
  be set to 1).
- **Slightly more memory in flight.** Up to `signWorkers + 1024 +
  1024` messages queued during a publish stall vs. `1 + 100`
  before. At ~3 KiB per signed message that's still under 10 MiB
  worst case.
- **One more goroutine pair on shutdown.** The shutdown sequence
  now closes `mqttPubCh`, sign workers drain and exit, the last one
  closes `mqttSignedCh`, the publish worker drains and exits. The
  existing `autopahoWg` covers all of this.

### Expected payoff

JWS signing parallelises across `GOMAXPROCS` cores instead of running
on one. The 14 % single-core CPU pin disappears from the profile;
signing throughput scales close to linearly with worker count up to
the point where the broker's accept rate (or the network) becomes
the cap. On the load-gen smoke this removes MQTT publishing as the
governor on EDM throughput — the test box now hits the next
bottleneck (network, parquet writers) rather than this one.

## Change 7: atomic.Pointer for ignore-set lookups

**File:** `pkg/runner/runner.go`. Affects struct, `setIgnoredClientIPs`,
`setIgnoredQuestionNames`, `clientIPIsIgnored`, `questionIsIgnored`,
`getNumIgnoredClientCIDRs`.

### Why this was a bottleneck

The hot path called `clientIPIsIgnored` and `questionIsIgnored` on
every frame. Both took an `RWMutex.RLock()` even when the underlying
ignore set was unset (the common production case). Reload writers
took `Lock()` on `setIgnored*`, which is rare but still imposed the
slow-path cost on every reader.

### What we changed

- `ignoredClientsIPSet *netipx.IPSet` and `ignoredQuestions dawg.Finder`
  become `atomic.Pointer[netipx.IPSet]` and `atomic.Pointer[dawgFinderHolder]`.
- `ignoredClientCIDRsParsed` becomes `atomic.Uint64`.
- `dawgFinderHolder` is a tiny wrapper struct containing `finder dawg.Finder`,
  needed because `atomic.Pointer` requires a concrete type and `dawg.Finder`
  is an interface.
- All RWMutex protect-everything-with-Lock blocks become
  single-line `atomic.Store` calls.
- The hot path is now:

  ```
  ipset := edm.ignoredClientsIPSet.Load()
  if ipset == nil { return false }
  ```

  Zero allocations, no lock, no contention.

### Tradeoff

- **The previous code called `dawg.Finder.Close()` on the old finder
  during reload.** With atomic-pointer swap we cannot safely close
  the old finder, because hot-path readers may still hold a pointer
  to it. So we *don't* close it — we just drop the reference and
  let GC reclaim it. The dawg implementation uses an mmap'd file
  internally; finalisers eventually unmap it, so the leak is bounded
  per-rotation. With reloads being rare (manual SIGHUP-style events),
  this is acceptable. If you ever need promptness, an
  epoch-based reclamation scheme would work, but the current load
  is so far from triggering a problem here that the simpler choice
  wins.
- The CIDR count getter `getNumIgnoredClientCIDRs` is now a pure
  `atomic.Uint64.Load()`. No semantic change.

### Expected payoff

The two RWMutexes leave the hot path. In mutex profiles the
`ignoredClientsIPSetMutex` and `ignoredQuestionsMutex` block-time
entries become invisible. Direct CPU savings are small (RLock/RUnlock
is fast in the uncontended case) but goroutine wakeups during
reload no longer ripple through every minimiser worker. Worth doing
mostly for cleanliness, with a small steady-state win.

## Change 8: WKD tracker — split read snapshot from map mutation

**File:** `pkg/runner/runner.go`. Affects `wellKnownDomainsTracker`,
`newWellKnownDomainsTracker`, `lookup`, `rotateTracker`, the
`wkd.dawgModTime` reference inside `dataCollector`, and the retryer's
log message.

### Why this was a bottleneck

`wkd.lookup(msg)` ran on every frame's hot path. It took
`wkd.mutex.RLock()` to read `dawgFinder` + `dawgModTime`. The same
lock was taken for `Lock()` by `rotateTracker` (a once-per-minute
event). The aggregator map `wkd.m` was *also* nominally protected
by this lock, but only one goroutine (`dataCollector`) ever wrote
to it, and no one else read it in the hot path — so the mutex was
mainly serialising the per-frame DAWG reads against a once-per-minute
write.

### What we changed

The struct gains:

```go
type wkdSnapshot struct {
    dawgFinder  dawg.Finder
    dawgModTime time.Time
}

type wellKnownDomainsTracker struct {
    snap atomic.Pointer[wkdSnapshot]
    m    map[int]*histogramData  // single-writer (dataCollector)
    ...
}
```

`lookup()` becomes:

```go
snap := wkd.snap.Load()
dawgIndex, suffixMatch := getDawgIndex(snap.dawgFinder, name)
return dawgIndex, suffixMatch, snap.dawgModTime
```

No lock. `rotateTracker` reads `wkd.snap.Load()` to compare the on-disk
modtime, builds a fresh `wkdSnapshot` if the file changed, atomically
swaps it in, and swaps the histogram map. Both swaps happen in the
same goroutine that writes to `m`, so no lock is needed for the map
either.

The `dataCollector` retryer's "discard stale update" check changes
from `wu.dawgModTime != wkd.dawgModTime` to
`wu.dawgModTime != wkd.snap.Load().dawgModTime`.

### Tradeoff

- **Slight torn-state risk during rotation.** A worker mid-frame that
  loads `snap` just before `rotateTracker` swaps it will use the old
  `dawgFinder` and the old `dawgModTime`. That's actually the *desired*
  behaviour: the update will arrive at `dataCollector` carrying the
  old `dawgModTime`, the modtime-mismatch check will fire, and the
  update goes onto `retryCh` for re-lookup against the new DAWG. So
  the existing retryer mechanism already handles this case correctly.
- **`rotateTracker` no longer touches the old `wkd.mutex`.** Anything
  outside this package that synchronised against that mutex would
  break — but the mutex was unexported, so nothing outside the
  package could touch it.
- **The old `wellKnownDomainsData.dawgFinder` field is still used**
  for the post-rotation parquet writeback (it carries the *previous*
  DAWG so name lookups for histogram output still work). The
  rotated `*wellKnownDomainsData` snapshot continues to embed
  `dawgFinder`; the change is only internal to the tracker.

### Expected payoff

Removes another RWMutex from every frame. The `wkd.mutex` block-time
entry disappears from the mutex profile. Direct CPU is small. The
real win is that all four Tier-2 changes together open the per-frame
hot path so that the only synchronisation a worker hits is the rate
limiter on the input channel and its own private LRU. Throughput
should now scale linearly with `MinimiserWorkers` until the input
channel itself becomes the bottleneck (a Tier-3 candidate).

## Verification

Unit tests pass after the refactor:

```bash
$ go test -count=1 -short ./pkg/runner
ok      github.com/dnstapir/edm/pkg/runner   0.28s
```

End-to-end with the load-gen at QPS=0,
MQTT publishing on, default minimiser worker count
(`--minimiser-workers 0` = GOMAXPROCS via Tier 1 #4), default
sign-worker count (= GOMAXPROCS via Tier 2 #6):

- pre-Tier-2 ceiling: ~140 K qps, EDM CPU ~6.5 cores
- post-Tier-2 ceiling: re-measure and update this section

The expected next ceiling is the input channel: a single
listener goroutine still funnels every frame through `inputChannel`.
With Tier-1's 1024 buffer the channel itself absorbs jitter, but
the listener can only deserialise one frame at a time. This is the
Tier-3 entry point ("per-worker pipeline" or "multi-listener input").

## Side effect: paho file queue exposure

Tier 2 lifts the JWS-sign rate from ~40 K/s (sequential, single goroutine)
to whatever GOMAXPROCS allows (~10×). The next bottleneck downstream
turned out to be EDM's default-on **MQTT file queue**, configured at
`runner.go:684` to `<DataDir>/mqtt/queue`. Paho's
`autopaho/queue/file.(*Queue).Peek` calls `oldestEntry()` on every
dequeue, which does:

```go
entries := os.ReadDir(dir)        // O(N)
for _, e := range entries {
    info := os.Lstat(...)         // O(N) syscalls + allocations
    // pick oldest
}
```

That is O(N) per dequeue, O(N²) per N enqueues. As the queue fills, the
manager's drain rate falls below the enqueue rate, the queue grows, GC
pressure climbs, and end-to-end throughput collapses. Symptom from the
outside: rate runs steady for some seconds, then plummets to near zero
as the queue crosses ~10 K files. We confirmed this on a stress run by
heap-profiling EDM under sustained 100 K qps:

```
55% alloc_space   autopaho.managePublishQueue → file.Queue.Peek
                  → oldestEntry → os.ReadDir + os.Lstat
```

…and observing the queue dir grow to **88 903 files / 350 MB** at the
moment of the stall. Goroutine count stayed flat at ~96 — this is not
a goroutine leak; it is allocation pressure compounding into GC stalls.

### Why this isn't a Tier 2 regression in the strict sense

Pre-Tier-2, JWS signing capped the rate at ~40 K/s, well below the
queue's collapse point. The queue's O(N²) behaviour was always there;
Tier 2 just removed the upstream throttle that was hiding it. A
deployment that has its broker available and processes load at any
non-trivial sustained rate will eventually hit this regardless of
Tier 2 — Tier 2 only made the time-to-failure noticeable in a 100 K-qps
load test.

### What we did

For dev/test setups (the load-gen smoke), the file queue is unnecessary:
the broker is always up, QoS 0 publishes are best-effort anyway, and
the durability the queue provides has no value. The fix is the existing
`--disable-mqtt-filequeue` flag (already in EDM's CLI surface).

Without the file queue, paho buffers messages in-memory only, with its
own bounded internal queue applying real backpressure. In our smoke at
100 K qps with all Tier 1 + Tier 2 changes:

- `edm_new_qname_discarded_total`: **0** (vs ~94 % discards before)
- queue dir: empty
- end-to-end throughput: stable ~67 K qps for the 3-minute run

### What we did *not* do

We did **not** change the file queue's default. The queue is meant to
provide durability across broker disconnects — disabling it weakens
that guarantee, and operators in real deployments should make that
choice consciously. The right long-term fix is paho-side: replace the
`oldestEntry()` directory scan with an in-memory ordered index of the
queue files (the same maintenance pass paho already does for cleanup
could maintain it). That is upstream work, not in scope here.

If you do run a deployment with the file queue enabled at high publish
rates, monitor `<DataDir>/mqtt/queue` size and pre-emptively trim or
disable when the broker can't keep up. Adding a queue-size Prometheus
metric would also be worth doing — open issue.

## Release Notes

* **New Features**
  * Added `--pebble-sync` flag to control write synchronization behavior.

* **Improvements**
  * Enhanced MQTT signing pipeline with configurable parallel workers.
  * Improved timestamp formatting in aggregation output with RFC3339 and ISO-8601 standards.
  * Performance optimizations for pseudonymization processing.

* **Bug Fixes**
  * Fixed handling of DNS messages with empty question sections.
  * Improved HTTP response cleanup and error management.

* **Chores**
  * Updated gitignore rules.


jschlyter:
To help code review, please keep PRs as small as possible, addressing a single feature or bug per request. Also make sure all commits are signed by the developer as required.

We do appreciate all contributions – we (and our bots) will carefully review your changes in smaller portions.

linkdata:
You're right. My apologies. I'll resubmit as requested.

PR #170, opened 2026-04-30
fix(config): allow optional TLS client CA

linkdata:
## Summary
- allow TLS listener configs to omit a client CA file
- keep cert/key validation intact when TLS input is enabled
- add focused validation coverage

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * TLS configuration no longer requires a client CA file to be specified.

* **Tests**
  * Added unit tests validating TLS configuration behavior without client CA requirements.

### DIFF
--- a/pkg/runner/config_validation_test.go
+++ b/pkg/runner/config_validation_test.go
@@ -1,0 +1,30 @@
+package runner
+
+import "testing"
+
+func TestConfigValidationAllowsTLSWithoutClientCA(t *testing.T) {
+	conf := validConfigForValidation()
+	conf.InputUnix = ""
+	conf.InputTLS = "127.0.0.1:53535"
+	conf.InputTLSCertFile = "server-cert.pem"
+	conf.InputTLSKeyFile = "server-key.pem"
+
+	if err := validate.Struct(conf); err != nil {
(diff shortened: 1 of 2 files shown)


linkdata:

## Optional TLS Client CA

- **Bug:** `input-tls-client-ca-file` was validated as required whenever `input-tls` was set, even though the runtime only enables client certificate authentication when the CA file is configured.
- **Impact:** Operators could not run a TLS dnstap listener without requiring client certificates.
- **Fix:** Removed the required validation from `input-tls-client-ca-file` while keeping the existing conditional mTLS setup in the listener.
- **Reasoning:** Server TLS and mTLS are separate choices; an omitted client CA should mean encrypted transport without client cert verification.
- **Tests:** Added config validation coverage showing `input-tls` with server cert/key and no client CA is valid.

linkdata:
Closing as already addressed on `main`. The identical change — dropping `validate:"required_with=InputTLS"` from the `InputTLSClientCAFile` config field so the client CA is optional — landed via #213 (commit `705820d`). Current `main` already declares the field without the `required_with` constraint (`pkg/runner/runner.go`), so this PR is now a no-op. Thanks!

PR #171, opened 2026-04-30
fix(protocols): tolerate empty question events

linkdata:
## Summary
- make NewQnameEvent safe for DNS messages with no questions
- preserve header-derived fields while leaving qname/qtype/qclass empty
- update the doc comment to match the permissive behavior

## Tests
- go test ./pkg/protocols ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when processing DNS messages that contain an empty question section; related event fields now default sensibly instead of causing errors.

* **Tests**
  * Added unit test coverage to verify safe handling of DNS messages with empty question sections and to ensure produced events contain expected metadata and sensible defaults.

### DIFF
--- a/pkg/protocols/protocols.go
+++ b/pkg/protocols/protocols.go
@@ -95,20 +95,28 @@ func bitsFromMsg(dns *dns.Msg) uint16 {
 	return bits
 }
 
+// NewQnameEvent constructs a NewQnameJSON event from a DNS message and a
+// timestamp. If the Question section is empty the returned event will have
+// an empty Qname, nil Qtype and nil Qclass, but other header-derived fields
+// will still be populated.
 func NewQnameEvent(msg *dns.Msg, ts time.Time) NewQnameJSON {
 	bits := bitsFromMsg(msg)
 	flags := int(bits)
 
-	qType := int(msg.Question[0].Qtype)
-	qClass := int(msg.Question[0].Qclass)
-
-	return NewQnameJSON{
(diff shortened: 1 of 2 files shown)


linkdata:

## NewQnameEvent Empty Question Section Panic

- **Bug:** In `pkg/protocols/protocols.go`, `NewQnameEvent()` accessed `msg.Question[0]` without first checking whether the Question section had any entries. While all current callers in the runner package validate this before calling, the function itself was unsafe and would panic on an empty Question section.
- **Impact:** A future caller (or test) that called `NewQnameEvent()` with an empty Question section would panic. As a public function in a published package, this is a defensive programming violation.
- **Fix:** Added a length check on `msg.Question`. If empty, the returned event has an empty `Qname` and nil `Qtype`/`Qclass` pointers, while header-derived fields (Flags, Type, Version) are still populated.
- **Reasoning:** Public functions should validate their inputs or be safe against edge cases. The fix provides graceful degradation rather than panicking.
- **Tests:** Added `TestNewQnameEventValid` and `TestNewQnameEventEmptyQuestion` in `pkg/protocols/protocols_test.go` to verify both the happy path and the empty-question case.

linkdata:
@coderabbitai review

linkdata:
@coderabbitai full review

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #172, opened 2026-04-30
fix(mqtt): normalize server URLs

linkdata:
## Summary
- accept bare MQTT host:port values by defaulting them to tls://
- reject unsupported schemes and missing hostnames, including tls://:8883
- add focused URL parsing tests

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * MQTT server configuration now validates URLs more strictly, rejecting unsupported schemes and malformed inputs.
  * Server addresses without an explicit scheme are now automatically treated as TLS connections.
  * Hostname validation now requires non-empty values.

* **Tests**
  * Added comprehensive unit tests for MQTT server URL parsing and validation.
  * Updated existing test to use valid MQTT server URL format.

### DIFF
--- a/pkg/runner/mqtt.go
+++ b/pkg/runner/mqtt.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/url"
+	"strings"
 	"sync"
 
 	"github.com/eclipse/paho.golang/autopaho"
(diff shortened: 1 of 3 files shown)


linkdata:

## MQTT Server URL Normalization

- **Bug:** The default MQTT server value was documented and configured as a bare `host:port`, but the MQTT client expects URL-shaped values.
- **Impact:** A default or operator-provided bare address could be parsed with the host portion as a scheme and fail later during connection setup.
- **Fix:** Bare MQTT server values are normalized to `tls://host:port`; explicit schemes are parsed and preserved.
- **Reasoning:** The default port is the TLS MQTT port, so the secure scheme is the least surprising default while still allowing explicit plaintext or websocket schemes.
- **Tests:** Added MQTT URL parsing coverage for bare IPv4/IPv6 addresses, explicit TLS/MQTT/TCP schemes, missing hosts, and unsupported schemes.

oej:
Do we have a list of URL schemas that EDM actually supports? The parse function includes everything possible on earth, but we may not want to support that.

linkdata:
For EDM, you could just decide what you want to support. 
The enforced list in the PR is fairly conservative - not everything the parse function accepts will be let through.

For MQTT, it's a bit fuzzy. MQTT.js supports mqtt, mqtts, tcp, tls, ws, wss, plus some platform-specific schemes like wxs and alis.

linkdata:
@coderabbitai review

linkdata:
@coderabbitai full review

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #174, opened 2026-04-30
fix(runner): improve timestamp filename errors

linkdata:
## Summary
- validate histogram filename shape before indexing timestamp parts
- return explicit parse errors for malformed start/stop timestamps
- add malformed filename regression coverage

## Tests
- go test ./pkg/runner ./...


linkdata:

PR #176, opened 2026-04-30
fix(aggregateSender): close response bodies on read errors

linkdata:
## Summary
- defer response body close immediately after aggregate sender receives a response
- keep the body closed even when reading the response fails
- add a truncated-response regression test

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced HTTP response handling to ensure response bodies are reliably closed and cleaned up even when errors occur during response processing.

### DIFF
--- a/pkg/runner/aggregate_sender.go
+++ b/pkg/runner/aggregate_sender.go
@@ -138,15 +138,11 @@ func (as aggregateSender) send(fileName string, ts time.Time, duration time.Dura
 	if err != nil {
 		return fmt.Errorf("sendAggregateFile: unable to send request, elapsed time %s: %w", elapsedTime, err)
 	}
+	defer res.Body.Close()
 
 	bodyData, err := io.ReadAll(res.Body)
 	if err != nil {
 		return fmt.Errorf("sendAggregateFile: unable to read response body: %w", err)
-	}
-
-	err = res.Body.Close()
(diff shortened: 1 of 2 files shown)


linkdata:

## HTTP Response Body Leak on Read Error

- **Bug:** In `aggregateSender.send()`, if `io.ReadAll(res.Body)` returned an error, the function returned early without closing `res.Body`. This leaked the HTTP connection from the transport pool.
- **Impact:** Transient network errors reading the aggregate receiver's response body permanently removed one HTTP connection from the pool, gradually exhausting the pool over time.
- **Fix:** Added `defer res.Body.Close()` immediately after the successful `Do()` call and removed the explicit `res.Body.Close()` call that was only reachable on the success path.
- **Reasoning:** Using `defer` ensures the body is closed on all return paths, including error paths, which is the idiomatic Go pattern for HTTP response bodies.
- **Tests:** Added `TestAggregateSenderClosesBodyOnReadError` which uses an HTTP connection hijacker to truncate the response body mid-delivery, verifying that `send()` returns an error without leaking the connection.

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #177, opened 2026-04-30
fix(runner): release fs watcher lock on removal errors

linkdata:
## Summary
- release cleanupFSWatchers RLock with defer on all exit paths
- add a small removal hook so the error path can be exercised directly
- assert the mutex is released after a watcher removal failure

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -1094,6 +1094,7 @@ func (edm *dnstapMinimiser) addFSWatchers(fileToFuncs map[string][]func() error)
 
 func (edm *dnstapMinimiser) cleanupFSWatchers() error {
 	edm.fsWatcherMutex.RLock()
+	defer edm.fsWatcherMutex.RUnlock()
 	for _, watchPath := range edm.fsWatcher.WatchList() {
 		watchPathInUse := false
 		for fsWatcherFuncFilename := range edm.fsWatcherFuncs {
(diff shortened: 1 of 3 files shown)


linkdata:

## FS Watcher Mutex Leak

- **Bug:** In `cleanupFSWatchers()`, an early `return` statement inside the `RLock` scope at lines 1003–1005 exited the function without calling `RUnlock()`, leaking the read lock.
- **Impact:** Any subsequent attempt to acquire the write lock on `fsWatcherMutex` (e.g., in `configureFSWatchers`) would block indefinitely because the read lock was never released.
- **Fix:** Added `defer edm.fsWatcherMutex.RUnlock()` immediately after `RLock()` and removed the explicit `RUnlock()` call that was only reachable on the success path.
- **Reasoning:** Using `defer` ensures the lock is always released, even on error return paths, preventing mutex deadlocks.
- **Tests:** Added `TestCleanupFSWatchersReleasesLockOnError` which calls `cleanupFSWatchers()` and then verifies `fsWatcherMutex.TryLock()` succeeds, confirming the read lock was released.

linkdata:
Superseded by #229 — combined with the PRs it's coupled to (must merge together) into a single PR to remove the merge-order rebase dance. All fixes from this PR are included there.

PR #178, opened 2026-04-30
fix(runner): skip malformed dnstap frames

linkdata:
## Summary
- skip malformed or incomplete dnstap frames instead of exiting the minimiser loop
- keep later valid frames processable after bad input
- register minimiser cleanup immediately in the regression test

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/run_minimiser_test.go
+++ b/pkg/runner/run_minimiser_test.go
@@ -1,0 +1,86 @@
+package runner
+
+import (
+	"io"
+	"log/slog"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/pebble"
+	dnstap "github.com/dnstap/golang-dnstap"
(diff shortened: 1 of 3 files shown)


linkdata:

## Malformed Dnstap Frame Handling

- **Bug:** A single protobuf unmarshal error caused a minimiser worker to exit, and missing message/type fields could panic before the frame was dropped.
- **Impact:** One malformed frame could permanently reduce processing capacity or crash the process.
- **Fix:** Malformed frames and structurally incomplete dnstap messages are logged and skipped while the worker continues.
- **Reasoning:** Capture streams can contain bad frames; one invalid payload should not terminate long-lived processing.
- **Tests:** Added a worker-level regression that sends malformed bytes, a dnstap message without payload, a payload without message type, and then a valid response that must still be processed.

linkdata:

PR #179, opened 2026-04-30
fix(runner): tolerate partial dnstap metadata

linkdata:
## Summary
- tolerate dnstap messages with missing timestamps and socket metadata
- avoid formatting empty addresses as invalid endpoints
- assert all IP-derived session fields remain nil when metadata is absent

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust DNSTAP handling: avoids panics, drops malformed/partially-populated frames with diagnostics, and consistently yields zero/UTC timestamps when metadata is missing while tolerating absent socket/protocol information.

* **Tests**
  * Added unit tests covering missing timestamps/messages, incomplete socket metadata, and endpoint-formatting variations.

### DIFF
--- a/pkg/runner/partial_dnstap_test.go
+++ b/pkg/runner/partial_dnstap_test.go
@@ -1,0 +1,104 @@
+package runner
+
+import (
+	"net/netip"
+	"testing"
+	"time"
+
+	dnstap "github.com/dnstap/golang-dnstap"
+	"github.com/miekg/dns"
+)
+
+func TestParsePacketMissingTimestamps(t *testing.T) {
+	edm := discardEDM()
+	wire := packedDNSMsg(t, "example.com.", dns.TypeA, dns.RcodeSuccess)
+	epoch := time.Unix(0, 0).UTC()
+
+	tests := []struct {
+		name    string
+		isQuery bool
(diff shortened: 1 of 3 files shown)


linkdata:

## Partial Dnstap Message Nil-Safety

- **Bug:** Some optional dnstap protobuf fields were dereferenced as required fields, and query endpoint formatting checked `ResponsePort` while dereferencing `QueryPort`.
- **Impact:** Decodable but partially populated dnstap messages could panic processing; endpoint logs could also crash when only a query port was present.
- **Fix:** Packet parsing now uses safe timestamp fallback and shared endpoint formatting, while session creation tolerates missing socket family/protocol metadata.
- **Reasoning:** Optional protobuf fields should produce dropped packets or partial output, not process panics.
- **Tests:** Added coverage for missing query/response timestamps, missing socket family/protocol in session creation, and query-port-without-address endpoint formatting.

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):
Just one of the reasons DNSTAP annoys me

PR #181, opened 2026-04-30
fix(runner): safely reload DAWG files during rotation

linkdata:
## Summary
- use loadDawgFile when rotating the well-known-domain tracker
- avoid panics from dawg.Load on zero-byte DAWG files
- add regression coverage for an empty file appearing during rotation

## Tests
- go test ./pkg/runner ./...


linkdata:

PR #182, opened 2026-04-30
fix(runner): deduplicate concurrent new qnames

linkdata:
## Summary
- serialize first-seen checks per qname with sharded locks
- prevent concurrent workers from publishing duplicate new-qname events for the same name
- add a concurrent qnameSeen regression test

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/run_minimiser_test.go
+++ b/pkg/runner/run_minimiser_test.go
@@ -1,0 +1,51 @@
+package runner
+
+import (
+	"sync"
+	"testing"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/miekg/dns"
+)
+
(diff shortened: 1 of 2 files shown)


linkdata:

## Concurrent New-Qname Deduplication

- **Bug:** The first-seen qname check used separate LRU and Pebble operations without an atomic per-qname critical section.
- **Impact:** Two minimiser workers processing the same previously unseen qname at the same time could both decide it was new and enqueue duplicate `new_qname` events.
- **Fix:** Added sharded per-qname locking around the LRU/Pebble check-add-set sequence.
- **Reasoning:** Identical qnames must be deduplicated as a single logical operation, while unrelated qnames should still proceed concurrently without a global lock.
- **Tests:** Added concurrent `qnameSeen` coverage proving exactly one worker reports a qname as first-seen.

linkdata:

PR #183, opened 2026-04-30
fix(runner): flush collector data on shutdown

linkdata:
## Summary
- drain pending session and histogram updates after the collector stop signal
- flush final session and histogram data before closing writer channels
- preserve the original interval start for shutdown flushes

## Tests
- go test ./pkg/runner ./...

## Release Notes

* **Tests**
  * Added test coverage verifying proper data flushing on shutdown.

* **Improvements**
  * Enhanced interval start time tracking for sessions and histograms to ensure accurate data synchronization during collection and shutdown.

### DIFF
--- a/pkg/runner/data_collector_test.go
+++ b/pkg/runner/data_collector_test.go
@@ -1,0 +1,156 @@
+package runner
+
+import (
+	"io"
+	"log/slog"
+	"net/netip"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/miekg/dns"
+	"github.com/smhanov/dawg"
+	"github.com/spaolacci/murmur3"
+)
+
+func TestDataCollectorFlushesPendingDataOnShutdown(t *testing.T) {
+	edm, wkdTracker := newDataCollectorTestFixture(t, "example.com.")
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go edm.dataCollector(&wg, wkdTracker, "unused-in-shutdown-test.dawg")
+
+	serverID := "serverID"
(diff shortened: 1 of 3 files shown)


linkdata:

## Shutdown Flush

- **Bug:** The collector only wrote sessions and histograms on the minute ticker; pending data could be left in memory when shutdown happened before the next tick.
- **Impact:** The final partial interval of session rows and well-known-domain histogram updates could be silently lost during graceful shutdown.
- **Fix:** After minimiser workers stop, the collector now drains queued session/update work and flushes any accumulated session and histogram data before closing writer channels.
- **Reasoning:** Graceful shutdown should preserve already accepted data even if the current collection interval is incomplete.
- **Tests:** Added collector shutdown coverage that queues a session and histogram update, stops the collector, and verifies both writer channels receive flushed data.

## Shutdown Partial Interval Metadata

- **Bug:** Shutdown-flushed session and histogram files still derived their start time as `rotationTime - 60s`, even when shutdown happened partway through the current interval.
- **Impact:** The final partial interval could be written with filenames and aggregate metadata that overlapped the previous minute or claimed a longer collection window than the data actually covered.
- **Fix:** The collector now tracks session and histogram interval starts explicitly and carries those starts into flushed writer payloads; legacy callers still fall back to the previous one-minute calculation.
- **Reasoning:** Once shutdown can flush partial intervals, the writer contract needs both interval boundaries instead of reconstructing the start from the stop time.
- **Tests:** Extended collector shutdown coverage to assert flushed session and histogram payloads carry non-zero, non-inverted interval boundaries.

linkdata:

PR #186, opened 2026-04-30
fix(runner): stop config updater on shutdown

linkdata:
## Summary
- make configUpdater exit when the minimiser context is cancelled
- buffer and non-block config change notifications to avoid watcher stalls
- close the notification channel during Run shutdown

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -616,13 +616,23 @@ func configUpdater(viperNotifyCh chan fsnotify.Event, edm *dnstapMinimiser) {
 	})
 	t.Stop()
 
-	for e = range viperNotifyCh {
-		// If an event has been recevied this means we now want to
-		// enable the timer so the function will be called "soon", but
-		// if more events occur we will reset it again. This allows us
-		// to wait until events on the file settles down before
-		// actually calling the update function.
-		t.Reset(configUpdateDebounce)
+	for {
+		select {
+		case event, ok := <-viperNotifyCh:
+			if !ok {
+				return
+			}
+			e = event
+			// If an event has been recevied this means we now want to
+			// enable the timer so the function will be called "soon", but
+			// if more events occur we will reset it again. This allows us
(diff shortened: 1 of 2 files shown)


linkdata:

## Config Updater Goroutine Leak

- **Bug:** In `Run()`, `configUpdater` looped over an unbuffered `viperNotifyCh` that was never closed. The `for e = range viperNotifyCh` loop at line 601 could only exit when the channel was closed, which never happened, causing the goroutine to leak on shutdown. Additionally, the `viper.OnConfigChange` callback did a blocking send on the unbuffered channel, which could deadlock the viper watcher goroutine if `configUpdater` was busy.
- **Impact:** The `configUpdater` goroutine leaked on every `Run()` call, accumulating unreleased resources in long-lived processes or test suites.
- **Fix:** Three changes: (1) Made `viperNotifyCh` buffered with capacity 1 (line 1134), (2) Changed the `viper.OnConfigChange` callback to use a non-blocking send with `select` and `default` (lines 1139–1141), (3) Changed `configUpdater`'s loop from `for e = range` to a `select` that also listens to `edm.ctx.Done()` (lines 601–617), (4) Added `close(viperNotifyCh)` at the end of `Run()` after `wg.Wait()` (line 1405).
- **Reasoning:** Config-change events are idempotent (the dedup timer already coalesces multiple events), so dropping duplicates is correct. The goroutine must respond to context cancellation for clean shutdown. Closing the channel signals the end of events and allows the `range` loop to exit.
- **Tests:** Added `TestConfigUpdaterExitsOnContextCancel` which starts `configUpdater` with a buffered channel and verifies it exits promptly (within 2 seconds) when the context is cancelled.

linkdata:
Superseded by #229 — combined with the PRs it's coupled to (must merge together) into a single PR to remove the merge-order rebase dance. All fixes from this PR are included there.

PR #187, opened 2026-04-30
fix(runner): unblock session send on cancellation

linkdata:
## Summary
- make runMinimiser session sends respect context cancellation
- prevent shutdown from hanging when sessionCollectorCh is full
- add a blocked-send cancellation regression test

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/run_minimiser_test.go
+++ b/pkg/runner/run_minimiser_test.go
@@ -1,0 +1,57 @@
+package runner
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	dnstap "github.com/dnstap/golang-dnstap"
+	"github.com/dnstapir/edm/pkg/protocols"
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/miekg/dns"
+)
+
+// TestRunMinimiserSessionSendUnblocksOnContextCancel verifies that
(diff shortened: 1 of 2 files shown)


linkdata:

## Minimiser Session Send Backpressure

- **Bug:** In `runMinimiser()`, the send to `sessionCollectorCh` at line 2121 was a plain blocking send. If the channel filled up (capacity 100), the minimiser goroutine would block inside the inputFrame processing case and could never reach the outer `case <-edm.ctx.Done()` to exit.
- **Impact:** A slow or stalled `dataCollector` could prevent minimiser workers from shutting down gracefully, blocking `minimiserWg.Wait()` and stalling the entire shutdown sequence including the histogram flush.
- **Fix:** Wrapped the channel send with a `select` that includes `case <-edm.ctx.Done()`, allowing the minimiser to exit when context is cancelled even if the channel is full.
- **Reasoning:** All sends inside the main minimiser select must have a context escape hatch to honour graceful shutdown.
- **Tests:** Added `TestRunMinimiserSessionSendUnblocksOnContextCancel` which fills `sessionCollectorCh` to capacity and verifies the minimiser exits promptly (within 2 seconds) when context is cancelled.

linkdata:

PR #189, opened 2026-04-30
fix(runner): track fs event watcher shutdown

linkdata:
## Summary
- track fsEventWatcher in the runner WaitGroup
- close the watcher before waiting so the goroutine exits cleanly
- keep the watcher pointer valid until after the goroutine has joined

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a file-watcher goroutine from lingering during shutdown, avoiding potential resource leaks and related errors.

* **Chores**
  * Tightened shutdown coordination so background watcher tasks are waited on and closed deterministically before exit.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -954,7 +954,8 @@ func (edm *dnstapMinimiser) getNumIgnoredClientCIDRs() uint64 {
 	return edm.ignoredClientCIDRsParsed
 }
 
-func (edm *dnstapMinimiser) fsEventWatcher() {
+func (edm *dnstapMinimiser) fsEventWatcher(wg *sync.WaitGroup) {
+	defer wg.Done()
 	// Like in
 	// https://github.com/fsnotify/fsnotify/blob/main/cmd/fsnotify/dedup.go
 	// we keep a timer per registered filename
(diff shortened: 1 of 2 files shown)


linkdata:

## fsEventWatcher Goroutine Not Tracked

- **Bug:** The `fsEventWatcher` goroutine was started at line 1204 with `go edm.fsEventWatcher()` and was not tracked in any WaitGroup. When `Run()` exited, the goroutine was forced to terminate abruptly via `fsWatcher.Close()` on the deferred path, with no way to verify it had completed.
- **Impact:** Inconsistent shutdown semantics; potential for fsEventWatcher to be processing an event while the rest of the program shuts down; in the worst case, a callback could run while shared state was being torn down.
- **Fix:** (1) Added a `*sync.WaitGroup` parameter to `fsEventWatcher` with `defer wg.Done()`. (2) Tracked it in the main `wg` via `wg.Add(1); go edm.fsEventWatcher(&wg)`. (3) Moved fsEventWatcher start to after `wg` is declared. (4) Added explicit `fsWatcher.Close()` before `wg.Wait()` so the goroutine has a signal to exit. (5) Made the deferred fsWatcher.Close into a safety net that checks for nil.
- **Reasoning:** All long-lived goroutines should be tracked by a WaitGroup for proper shutdown sequencing, matching the pattern used by other goroutines in the program.
- **Tests:** Verified by running the full test suite; no regressions.

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #190, opened 2026-04-30
fix(runner): gracefully shutdown HTTP servers

linkdata:
## Summary
- track pprof and metrics server goroutines in the runner WaitGroup
- gracefully shut down HTTP servers before waiting for worker exit
- leave dnstap ReadInto fire-and-forget because the library provides no stop hook

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Refactor**
  * Improved graceful shutdown and server cleanup to prevent hang-ups, ensure HTTP services stop cleanly, and surface only relevant shutdown errors in logs for clearer diagnostics.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -1322,19 +1322,52 @@ func run(edm *dnstapMinimiser, logger *slog.Logger, loggerLevel *slog.LevelVar) 
 		return fmt.Errorf("unable to create seen-qname LRU: %w", err)
 	}
 
+	var wg sync.WaitGroup
+
+	// The HTTP servers are tracked in their own WaitGroup so the deferred
+	// cleanup below can stop them and wait for their goroutines on every
+	// return path, including the early error returns during setup.
+	var serverWg sync.WaitGroup
+
 	pprofServer := newPprofServer(pprofListenAddr)
+	serverWg.Add(1)
 	go func() {
+		defer serverWg.Done()
 		err := listenAndServeHTTP(pprofServer)
-		logger.Error("pprofServer failed", "error", err)
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.Error("pprofServer error", "error", err)
+		}
 	}()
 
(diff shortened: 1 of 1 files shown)


linkdata:

## DNStap Reader Goroutine Limitation

- **Note:** The dnstap reader goroutine (`dti.ReadInto(edm.inputChannel)`) is started fire-and-forget without WaitGroup tracking. An earlier attempt to track it in `wg` and close `inputChannel` on shutdown caused SIGINT to hang because the upstream `golang-dnstap` library does not expose a stop/close mechanism for `ReadInto` — the goroutine blocks reading from the socket indefinitely. Closing `inputChannel` from `Run()` would also race with `ReadInto` sending and panic.
- **Impact:** The `ReadInto` goroutine leaks on shutdown, but the OS reclaims it on process exit.
- **Status:** Documented as an upstream library limitation. Fixing this properly would require either a patch to `golang-dnstap` to support cancellation, or closing the underlying listener from outside, which the current `FrameStreamSockInput` does not expose.

## HTTP Servers (pprof and metrics) Not Synchronized or Gracefully Shut Down

- **Bug:** The pprof and metrics HTTP servers (lines 1269-1272 and 1285-1288) were started as goroutines without WaitGroup tracking. Neither server received graceful shutdown signals; both `ListenAndServe()` blocks indefinitely. When `wg.Wait()` completed, they were still running.
- **Impact:** Unclean HTTP server shutdown; potential connection leaks; incomplete request handling; goroutines forced to terminate abruptly on process exit.
- **Fix:** (1) Wrapped both servers in tracked goroutines with `wg.Add(1)` and `defer wg.Done()`, handling `http.ErrServerClosed` as expected error. (2) Added graceful `Shutdown(ctx)` calls with 5-second timeout before `wg.Wait()`.
- **Reasoning:** HTTP servers need explicit graceful shutdown to drain pending requests and close connections cleanly; all goroutines should be tracked for proper ordering.
- **Tests:** Verified by running full test suite and checking that shutdown completes without errors.

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #175, opened 2026-04-30
fix(aggregateSender): use cancellable dialer

linkdata:
## Summary
- switch aggregate sender transport from Dial to DialContext
- preserve existing timeout and keepalive settings while allowing request cancellation to propagate

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Refactor**
  * HTTP sending now respects request contexts so connection setup and POST requests are cancellable and tied to operation lifecycle.
* **Behavior**
  * Send operations are now bound to shutdown/cancellation contexts to avoid hanging or orphaned requests during shutdown.
* **Tests**
  * Tests updated to exercise the context-aware send behavior.

PR #191, opened 2026-04-30
fix(aggregateSender): close transport on reload

linkdata:
## Summary
- retain the aggregate sender HTTP transport so it can be closed explicitly
- close the previous idle connection pool when aggregate sender config reloads
- avoid leaking reusable HTTP connections across reloads

## Tests
- go test ./pkg/runner ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP connection management during sender reloads: old transport idle keep-alive connections are now closed after a successful swap, and reloads preserve the existing sender on failure to avoid interruptions.

* **Tests**
  * Added a test that verifies idle HTTP connections are closed when the sender is reloaded, ensuring connection pooling is correctly released.

### DIFF
--- a/pkg/runner/aggregate_sender.go
+++ b/pkg/runner/aggregate_sender.go
@@ -27,6 +27,7 @@ type aggregateSender struct {
 	aggrecURL         *url.URL
 	caCertPool        *x509.CertPool
 	signingHTTPClient *httpsign.Client
+	httpTransport     *http.Transport
 }
 
 func (edm *dnstapMinimiser) newAggregateSender(aggrecURL *url.URL, signingJwk jwk.Key, caCertPool *x509.CertPool) (aggregateSender, error) {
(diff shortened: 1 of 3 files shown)


linkdata:

## HTTP Client Transport Resource Leak on Reload

- **Bug:** In `setupHistogramSender()`, when the aggregateSender was replaced (e.g., during config reload), the old aggregateSender's HTTP transport's connection pool was never closed. The transport holds idle connections that accumulate over time.
- **Impact:** Over multiple config reloads, idle HTTP connections accumulate and are not returned to the OS, exhausting per-process connection limits and file descriptors.
- **Fix:** (1) Added `httpTransport` field to `aggregateSender` struct to track the transport. (2) In `setupHistogramSender()`, call `CloseIdleConnections()` on the old transport before replacing it.
- **Reasoning:** HTTP transports maintain connection pools that must be explicitly cleaned up when the client is replaced, or connections will remain open until GC runs or the process exits.
- **Tests:** Verified by running full test suite with no regressions.

linkdata:
@coderabbitai review

linkdata:
⚠️ **Merge-order note** (automated conflict analysis vs the other open `ai` PRs)

This PR overlaps with **#175** in `pkg/runner/aggregate_sender.go` (`newAggregateSender`), so please **merge #175 first** — this branch will then need a small rebase (mostly `aggregate_sender.go`). Conflict-minimising order for the whole set:
`171 172 184 188 194 197 · 175→191 · 176→185 · 179→180 · 190→189 · 173→177→186 · 193→182→187→178 · 174→181→183→195`

linkdata (commented):
LGTM, this will have merge conflict since previous one added ContextDialer

morkrost (approved):

PR #192, opened 2026-04-30
fix(runner): add minimiser close cleanup

linkdata:
## Summary
- add dnstapMinimiser.Close for tests and non-Run callers
- cancel the minimiser context and close the fsnotify watcher safely
- register cleanup in tests that construct minimisers directly

## Tests
- go test ./pkg/runner ./...

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -1620,6 +1620,26 @@ func newDnstapMinimiser(logger *slog.Logger, edmConf edmConfiger) (*dnstapMinimi
 	return edm, nil
 }
 
+// Close releases resources held by the minimiser that aren't owned by the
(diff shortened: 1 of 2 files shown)


linkdata:
Closing as already addressed on `main`. The goal here — releasing the fsWatcher inotify instance and cancelling the stop context so repeated test runs don't exhaust `max_user_instances` — is already handled: the `newTestDnstapMinimiser` helper (`pkg/runner/runner_test.go`) registers a `t.Cleanup` that calls `edm.stop()` and `edm.fsWatcher.Close()` at every construction site, and `Run()` already does `defer edm.stop()` + `defer edm.fsWatcher.Close()`. There are no non-Run callers that would need the proposed exported `Close()` method, so it has no remaining purpose. Thanks!

PR #193, opened 2026-04-30
feat(runner): add pebble sync switch

linkdata:
## Summary
- add --pebble-sync / pebble-sync config support
- use NoSync for seen-qname writes by default and allow operators to opt into Sync
- add write option coverage

## Tests
- go test ./pkg/cmd ./pkg/runner ./...

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -35,6 +35,7 @@ func init() {
 	runCmd.Flags().Bool("disable-mqtt", false, "disable MQTT message sending")
 	runCmd.Flags().Bool("disable-mqtt-filequeue", false, "disable MQTT file based queue")
 	runCmd.Flags().Bool("enable-manual-parquet-rotation", false, "enable localhost HTTP endpoint for manually rotating session and histogram parquet files")
+	runCmd.Flags().Bool("pebble-sync", false, "fsync seen-qname pebble writes")
 
 	runCmd.Flags().String("input-unix", "", "create unix socket for reading dnstap (e.g. /var/lib/unbound/dnstap.sock)")
 	runCmd.Flags().String("input-tcp", "", "create TCP socket for reading dnstap (e.g. '127.0.0.1:53535')")
--- a/pkg/runner/run_minimiser_test.go
+++ b/pkg/runner/run_minimiser_test.go
@@ -1,0 +1,17 @@
+package runner
+
+import (
+	"testing"
+
+	"github.com/cockroachdb/pebble"
+)
+
+func TestSeenQnameWriteOptions(t *testing.T) {
+	if got := seenQnameWriteOptions(config{}); got != pebble.NoSync {
+		t.Fatalf("default seen-qname write option = %p, want %p", got, pebble.NoSync)
+	}
+
+	if got := seenQnameWriteOptions(config{PebbleSync: true}); got != pebble.Sync {
+		t.Fatalf("pebble-sync seen-qname write option = %p, want %p", got, pebble.Sync)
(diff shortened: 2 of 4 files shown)


linkdata:

## Change 1: drop `pebble.Sync` on the seen-qname write

**File:** `pkg/runner/runner.go`, the `qnameSeen` path (~line 1859).

**Before:**

```go
if err := pdb.Set([]byte(qname), []byte{}, pebble.Sync); err != nil {
```

**After:**

```go
if err := pdb.Set([]byte(qname), []byte{}, pebble.NoSync); err != nil {
```

### Why this was a bottleneck

Pebble's `WriteOptions.Sync = true` means *the WAL is fsynced before the
call returns*. There is no automatic batching, no time-based or
count-based grouping; every Sync write is its own fsync syscall (group
commit shares fsyncs across *concurrent* goroutines, but our minimiser
workers issue their writes sequentially per worker, so coalescing is
limited).

At a steady 200 K qps with ~10 % novel domains, that is **20 000 fsyncs
per second** to a single Pebble WAL. Even on fast NVMe (≈ 50 µs per
fsync) that is ~ 1 second of fsync wall-clock time per second — i.e.
total saturation of the WAL writer. Worse, while the pipeline is
fsync-bound, the input channel fills, the framestream socket buffer
fills, and our load-generator stalls in `pollDesc.waitWrite`. This
behaviour shows up in CPU profiles as low `runMinimiser` utilization
without any single hot function — the cores are *waiting*.

### What we changed

`pebble.NoSync` — the write is committed to the memtable and appended to
the WAL OS buffer, but does not block on `fsync`. The WAL is still
flushed periodically by Pebble itself (see `Options.WALBytesPerSync`,
default 0 = none, and the implicit flush on memtable rotation), and on
graceful shutdown.

### Tradeoff

The seen-qname store is a **deduplication hint, not a system of record**.
Its only purpose is to suppress duplicate `new_qname` MQTT publications
across restarts. With this change:

- on a clean shutdown nothing is lost (Pebble flushes its WAL on
  `db.Close`);
- on a hard crash you lose **at most a few seconds** of "we have already
  seen this qname" state;
- the consequence is that, on next startup, EDM re-publishes those few
  seconds of qnames as `new_qname` events to MQTT — which is bounded
  (one event per affected name) and self-correcting (subsequent
  observations land in the LRU/Pebble store and dedupe normally).

There is no correctness consequence for downstream parquet output,
histograms, or cardinality counts — those don't read the seen-qname
store. The only observable effect is a small blip of `new_qname` MQTT
events after a hard crash.

### Expected payoff

Removes the per-write fsync entirely. In our load tests this lifts the
MQTT-on ceiling from ~110 K to ~150 K+ qps and makes the system robust
under bursty traffic (no more multi-millisecond input-channel stalls
tied to fsync latency).

### If you want to keep durability semantics

A more conservative alternative — not taken in this fork, but suggested
upstream — is to keep `pebble.Sync` and instead pass these options to
`pebble.Open`:

```go
&pebble.Options{
    WALMinSyncInterval: func() time.Duration { return 500 * time.Microsecond },
    WALBytesPerSync:    1 << 20,
}
```

`WALMinSyncInterval` introduces a 500 µs coalescing window so that
concurrent novel-qname writes from N minimiser workers share a single
fsync. Pebble's `commitPipeline.syncLoop` does the actual group commit;
the option just widens its waiting window. Net throughput is similar and
durability is preserved at the cost of ~ 500 µs of added publish latency
per affected qname. Worth considering for an upstream PR.

---

linkdata:

PR #197, opened 2026-04-30
fix(config): isolate environment overrides

linkdata:
## Summary
- apply the configured Viper environment prefix before AutomaticEnv
- keep unprefixed environment variables from overriding config keys
- add prefixed and unprefixed env override coverage

## Tests
- go test ./pkg/cmd ./...

## Summary by CodeRabbit

* **Improvements**
  * Environment variables must use the DNSTAPIR_EDM_ prefix to override app configuration.
  * Hyphenated configuration keys in env vars are automatically mapped to underscores for matching.

* **Tests**
  * Added tests to verify prefixed env var handling, that unprefixed vars are ignored, and hyphen-to-underscore overrides work.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"log/slog"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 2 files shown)


linkdata:

## Config Env Isolation

- **Bug:** Viper accepted unprefixed environment variables, so unrelated ambient variables such as `DEBUG=release` were treated as EDM config.
- **Impact:** Startup could fail during config unmarshalling before any runtime work began, even when the config file and flags were valid.
- **Fix:** Environment overrides now require the `DNSTAPIR_EDM_` prefix and map hyphenated config keys to underscore-separated environment keys.
- **Reasoning:** Prefixed environment variables preserve intentional env-based config while preventing collisions with common shell or build environment variables.
- **Tests:** Added command-package tests proving `DEBUG=release` is ignored and `DNSTAPIR_EDM_DEBUG=true` still overrides `debug = false`.

pkg/cmd/root.go
```diff
@@ -77,3 +80,9 @@ func initConfig() {
 	// Make it so we can detect changes to the cryptopan secret in the config
 	viper.WatchConfig()
 }
+
+func configureEnv() {
+	viper.SetEnvPrefix(envPrefix)
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
```
oej: Do we need to add a new dependency for just this use? Or are we going to use "strings" in other parts too?

oej (commented):
Do we want tests in the same directory as the source or should we have a specific directory for tests?

Is this test something we want to automate to make sure there will be no regressions?

linkdata:
> Do we want tests in the same directory as the source or should we have a specific directory for tests?

Go tests are always in the same directory as the source they test. https://pkg.go.dev/testing

> Is this test something we want to automate to make sure there will be no regressions?

All tests are usually automated by running `go test` as part of CI/CD.

pkg/cmd/root.go
linkdata: `strings` is part of the standard library, and used all over the place.

oej: OK, thanks for the explanation.

oej:
OK @linkdata - thank you for your patience with me :-)

linkdata:
@coderabbitai review

linkdata (commented):
LGTM

morkrost (approved):

PR #210, opened 2026-06-01
[codex] Build container only on published releases

linkdata:
## Summary
- Remove the container workflow_run trigger for completed Build runs on main
- Drop the now-redundant job-level event guard so the workflow only runs for published releases

## Validation
- git diff --cached --check
- Ruby YAML parse of .github/workflows/container.yml

## Summary by CodeRabbit

* **Chores**
  * Simplified CI/CD workflow configuration to improve deployment process efficiency.

**Note:** This change is internal infrastructure-focused and does not impact end-user features or functionality.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,22 +1,12 @@
 name: Build container
 
 on:
-  workflow_run:
-    workflows:
-      - "Build"
-    branches:
-      - "main"
-    types:
-      - completed
   release:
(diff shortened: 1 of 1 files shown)


linkdata (commented):
Turns off building packages for all builds, only published ones.

morkrost (approved):
Seems reasonable. Let's only auto-build releases.

PR #208, opened 2026-06-01
test(runner): file-op seams + cover writer error branches (coverage stage 3)

linkdata:
## What

Stage 3 of the staged coverage effort.

- Add filesystem seams to `seams.go`: `osCreate`, `osRename`, `osRemove`, `osMkdirAll`, `osReadDir`, `osStat` (each wraps the matching `os` function).
- Route the file writers, renamer and directory scanners through them: `createFile`, `renameFile`, `createSessionFile`/`createHistogramFile` cleanup, `diskCleaner`, `histogramSender`. (`setupMQTT`'s `MkdirAll` is intentionally left to its own change to avoid overlap.)

This is a mechanical, behavior-preserving refactor: each seam defaults to the real `os` function. It unblocks injecting filesystem failures in tests — used here for the writers, and by the upcoming `writeRotatedParquet` (stage 4) and `diskCleaner`/`histogramSender` (stage 9) test work.

## Tests

New `pkg/runner/fileops_test.go`:
- `TestCreateFile` — missing-parent-dir retry, mkdir failure, non-ENOENT create error.
- `TestRenameFile` — missing-dest-dir retry, dir-exists-but-rename-fails, stat error, mkdir failure, non-ENOENT rename error.
- `TestSessionWriterLogsCreateError` / `TestHistogramWriterLogsCreateError` — the worker error-log branch (failure injected via `osCreate`).

## Coverage

- `createFile` 91.7% → **100%**, `renameFile` 80% → **100%**, `sessionWriter`/`histogramWriter` 57.1% → **100%**.
- `pkg/runner` 77.4% → 78.3%.

## Verification

`go test -race ./pkg/runner/` green (existing tests unchanged); `golangci-lint run ./pkg/runner/` reports 0 issues.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for file-operation error handling, retries, and worker error-logging to ensure robust behavior under filesystem failures.

* **Refactor**
  * Replaced direct filesystem calls with abstraction layers for directory and file operations to improve maintainability and testability.

PR #211, opened 2026-06-02
refactor(runner): extract writeRotatedParquet helper (coverage stage 4)

linkdata:
## What

Stage 4 of the staged coverage effort — the refactor half. `createSessionFile` and
`createHistogramFile` were ~60-line near-duplicates of the same
create → write → close → rename → cleanup-on-failure flow. Lift the shared logic into a
new helper:

```go
writeRotatedParquet(label, tmpName, finalName string, write func(io.Writer) error) (string, error)
```

owning `createFile`, the `fileOpen`/`writeFailed` cleanup defer (double-close avoidance
+ remove-on-failure), `Close`, and `osRename`. The two callers become thin wrappers that
supply the `write` closure (`writeSessionParquet` or `writeHistogramParquet`) and forward
errors with their own prefix so the returned error chain stays **byte-identical** to before.

Per the plan note ("medium-risk; review in isolation"), this PR is refactor-only. A follow-up
test PR adds `TestWriteRotatedParquet` (happy / createFile-error / write-error-cleanup /
Close-error / osRename-error) to bring the new helper to ~100%.

## Behavior preservation

- `fileOpen=false`-before-rename ordering preserved exactly; the defer skips `Close`
  after the explicit `Close`, and only `Remove`s when `writeFailed`.
- Rename failure still leaves the temp file dangling (matches original behavior).
- Error wrap chains `createSessionFile: ...` / `createHistogramFile: ...` preserved via
  caller-side `fmt.Errorf`.

## Two intentional log-line normalizations

- `"writing out session parquet file"` → `"writing out session file"`. The histogram
  path already used `"writing out histogram file"`; one form for both.
- The redundant `Error("sessionWriter")` / `Error("histogramWriter")` log that fired
  **inside** `createXFile` on write failure is dropped — the outer
  `sessionWriter`/`histogramWriter` worker already logs the same error one frame up.
  The only behavior change is one fewer duplicate log line on write-error.
  `TestSessionWriterLogsCreateError` / `TestHistogramWriterLogsCreateError` (introduced
  in #208) pass unchanged because they exercise the `osCreate`-failure path and still
  observe the outer log.

## Coverage

- `createSessionFile`: 60% → **100%**
- `createHistogramFile`: 59% → **100%**
- new `writeRotatedParquet`: 61.5% (rises in the follow-up test PR)
- `pkg/runner`: 78.3% → **79.5%**

## Verification

`go test -race ./pkg/runner/` green; `golangci-lint run ./...` reports 0 issues;
`gofmt -l` clean.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated internal parquet file rotation logic to ensure consistent handling between session and histogram outputs.
  * Improved error reporting and file lifecycle management across parquet write operations, reducing potential failure modes.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -2202,56 +2202,14 @@ func (edm *dnstapMinimiser) createSessionFile(ps *prevSessions, dataDir string) 
 	absoluteTmpFileName, absoluteFileName := buildParquetFilenames(sessionsDir, "dns_session_block", startTime, ps.rotationTime)
 
 	absoluteTmpFileName = filepath.Clean(absoluteTmpFileName) // Make gosec happy
-	edm.log.Info("writing out session parquet file", "filename", absoluteTmpFileName)
 
-	outFile, err := edm.createFile(absoluteTmpFileName)
+	name, err := edm.writeRotatedParquet("session", absoluteTmpFileName, absoluteFileName, func(w io.Writer) error {
+		return edm.writeSessionParquet(w, ps)
+	})
 	if err != nil {
-		return "", fmt.Errorf("createSessionFile: unable to open session file: %w", err)
+		return "", fmt.Errorf("createSessionFile: %w", err)
 	}
-	fileOpen := true
-	writeFailed := false
-	defer func() {
-		// Closing a *os.File twice returns an error, so only do it if
-		// we have not already tried to close it.
-		if fileOpen {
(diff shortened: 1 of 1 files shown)


linkdata (commented):
LGTM

morkrost (commented):
Ok

morkrost (approved):

PR #212, opened 2026-06-02
test(runner): cover writeRotatedParquet branches (coverage stage 4 tests)

linkdata:
## What

The follow-up test PR for [#211](https://github.com/dnstapir/edm/pull/211). Brings the
new `writeRotatedParquet` helper to 100% by covering every branch via the file-op seams
introduced in #208 plus a Close-injection trick that needs no new seams.

**Stacked on #211** — base is `coverage/stage-4-write-rotated-parquet`. Once #211 lands,
the base will retarget to `main` and the diff becomes test-only.

## Cases

- `happy` — file is created, written, renamed; temp gone, final has data
- `createFile error` — `osCreate` seam injection surfaces the error
- `write error removes temp file`
- `write error and remove failure is logged` — covers the defer's `osRemove`-failure branch
- `explicit Close error after successful write` — closure closes `outFile`, so the helper's
  explicit `Close()` hits `os.ErrClosed`, sets `writeFailed=true`, and the defer removes the temp
- `deferred Close error is logged` — closure closes `outFile` **and** returns an error so the
  defer's `Close()` runs against an already-closed handle
- `rename error leaves temp file` — documents preserved behavior

## How the Close-error injection works

Without faking `*os.File` (which would require a new seam), the only way to deterministically
fail `Close` on a healthy real file is to close it ourselves first. The write closure receives
the underlying `*os.File` as `io.Writer` and type-asserts back to call `Close()`. The next
`Close()` call (explicit or deferred) then returns `os.ErrClosed`. The asserts use `errors.Is`
so they survive any future error-chain wrapping.

## Coverage

- `writeRotatedParquet`: 61.5% → **100%**
- `pkg/runner`: 79.5% → **80.1%**

## Verification

`go test -race -count=2 ./pkg/runner/ -run TestWriteRotatedParquet` — all 7 subtests pass;
full `go test -race ./pkg/runner/` green; `golangci-lint run ./...` reports 0 issues.

### DIFF
--- a/pkg/runner/fileops_test.go
+++ b/pkg/runner/fileops_test.go
@@ -133,6 +133,158 @@ func TestRenameFile(t *testing.T) {
 	})
 }
 
+func TestWriteRotatedParquet(t *testing.T) {
+	t.Run("happy", func(t *testing.T) {
+		edm := discardEDM()
+		dir := t.TempDir()
+		tmp := filepath.Join(dir, "data.tmp")
+		final := filepath.Join(dir, "data")
+
+		name, err := edm.writeRotatedParquet("test", tmp, final, func(w io.Writer) error {
+			_, err := w.Write([]byte("hello"))
+			return err
+		})
+		if err != nil {
+			t.Fatalf("writeRotatedParquet: %v", err)
+		}
+		if name != final {
+			t.Fatalf("name = %q, want %q", name, final)
+		}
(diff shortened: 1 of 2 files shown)


linkdata (commented):
LGTM

linkdata:
@coderabbitai review

morkrost (approved):

PR #220, opened 2026-06-02
test(runner): cover pseudonymiseIP cache branches (coverage stage 9)

linkdata:
## What

Second slice of stage 9 (test-only backfill). Adds \`TestPseudonymiseIPCacheBranches\`
covering the three cache arms of \`pseudonymiseIP\` that
\`TestIPConversionErrorsAndPseudonymiseInvalid\` (bad-slice only) and
\`TestPseudonymiseDnstap\` (one-shot success on a fresh cache) do not reach:

- **cache hit on repeat** — second call against the same address takes the
  \`cryptopanCache.Get(addr)\` hit arm and the \`promCryptopanCacheHit.Inc()\` path.
- **cache eviction at size limit** — \`setCryptopan(_, _, 1)\` shrinks the LRU to a single
  entry; a second distinct address forces eviction, exercising the
  evicted/\`promCryptopanCacheEvicted.Inc()\` path.
- **cache disabled bypasses cache logic** — \`setCryptopan(_, _, 0)\` leaves
  \`cryptopanCache\` nil so the cache-Get and cache-Add branches are skipped entirely.

## Coverage

- \`pseudonymiseIP\`: 88.9% → **94.4%**

The residual 5.6% is the "\`Anonymize\` returned a slice that fails \`AddrFromSlice\`" arm,
which requires a contrived cryptopan implementation and is left as
accepted-as-uncovered.

## Verification

\`go test -race -count=2 ./pkg/runner/\` green; \`golangci-lint run ./...\` reports 0 issues.

## Summary by CodeRabbit

* **Tests**
  * Added tests covering IP pseudonymisation cache behavior: repeated pseudonymisation yields consistent output and a cache hit; cache size of 1 triggers eviction when a second address is processed; disabling cache keeps no cache but still produces pseudonymised output.

### DIFF
--- a/pkg/runner/runner_extra_test.go
+++ b/pkg/runner/runner_extra_test.go
@@ -451,6 +451,67 @@ func TestSetIgnoredQuestionNamesBranches(t *testing.T) {
 	})
 }
 
+// TestPseudonymiseIPCacheBranches covers the three pseudonymiseIP cache
+// branches that TestIPConversionErrorsAndPseudonymiseInvalid (bad-slice)
+// and TestPseudonymiseDnstap (one-shot success) do not reach: cache hit,
+// cache eviction at the LRU size limit, and the cache-disabled path
+// reached via cryptopanCache == nil.
+func TestPseudonymiseIPCacheBranches(t *testing.T) {
+	addrA := netip.MustParseAddr("198.51.100.20").AsSlice()
+	addrB := netip.MustParseAddr("198.51.100.30").AsSlice()
+
+	t.Run("cache hit on repeat", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


linkdata:
Addressed in 1b9b831: tightened the cache-hit assertion to check `cryptopanCache.Len() == 1` and `Contains(addrA)` — pinning the test to the observable side-effect of the hit arm rather than relying on the deterministic equality which would pass even if caching was bypassed.

linkdata:
@coderabbit review

morkrost (approved):
I admit, I had to check. The second example network doesn't get nearly as much use as 192.0.2.0/24. Thanks for reminding me.

PR #221, opened 2026-06-03
test(runner): cover manualParquetRotationHandler error/cancel branches (coverage stage 9)

linkdata:
## What

Third slice of stage 9 (test-only backfill). Adds three subtests for
\`manualParquetRotationHandler\` covering the branches that
\`TestManualParquetRotationHandlerRejectsNonPost\` (405) and
\`TestManualParquetRotationHandlerAcceptsCompletedRotation\` (202) do not reach:

- **Surfaces rotation error** — worker sends \`errInjected\` on \`req.done\`; handler
  responds 500 with the error string in the body.
- **Shuts down during send** — pre-fill \`parquetRotationRequestCh\` (capacity 1) to
  block the send-case of the first select, then cancel \`edm.ctx\`; handler takes the
  \`ctx.Done\` arm and responds 503 with "shutting down".
- **Cancel during wait** — worker consumes the request but withholds \`req.done\`;
  cancelling the HTTP request's context unblocks the second select's \`r.Context.Done\`
  arm; handler returns without writing, leaving the recorder at its default 200.

## Coverage

- \`manualParquetRotationHandler\`: 55.6% → **83.3%**
- \`pkg/runner\`: 84.5% → **84.9%**

## Residual (accepted-as-uncovered)

- The 504 \`time.After(manualParquetRotationWaitTimeout)\` path — needs \`synctest\` or
  a seam, since the constant is 9s.
- The \`r.Context().Done()\` arm of the **first** select — requires racing the request
  cancellation against the buffered channel send, which is hard to make deterministic
  without extra plumbing.

## Verification

\`go test -race -count=2 ./pkg/runner/ -run TestManualParquetRotationHandler\` green;
full \`go test -race ./pkg/runner/\` green; \`golangci-lint run ./...\` reports 0 issues.

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for the parquet rotation handler to verify error handling, graceful shutdown behavior, and request cancellation scenarios.

### DIFF
--- a/pkg/runner/manual_parquet_rotation_test.go
+++ b/pkg/runner/manual_parquet_rotation_test.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"context"
 	"io"
 	"log/slog"
 	"net/http"
@@ -8,6 +9,7 @@ import (
 	"net/netip"
 	"path/filepath"
 	"slices"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


linkdata (commented):
LGTM

morkrost (approved):

PR #173, opened 2026-04-30
fix(runner): retain sent histograms for 24 hours

linkdata:
## Summary
- correct sent histogram cleanup retention from 12 hours to 24 hours
- centralize the expiry check in a named helper used by diskCleaner
- test both sides of the 24-hour threshold

## Tests
- go test ./pkg/runner ./...

PR #229, opened 2026-06-08
fix(runner): disk-cleaner retention + fs-watcher lock + config-updater shutdown

linkdata:
Combines three coupled `runner` fixes that touch overlapping code (so they must merge together) into one PR — supersedes #173, #177, #186.

- **#173** — retain sent histograms for a true 24h (`sentHistogramRetention` const + `sentHistogramExpired` helper in `diskCleaner`).
- **#177** — `cleanupFSWatchers` releases the `fsWatcherMutex` RLock via `defer` on the removal-error path (was leaked).
- **#186** — config updater stops on shutdown: `select` on `ctx.Done()`, buffered `viperNotifyCh` + non-blocking `OnConfigChange` send.

Each fix is a separate commit. Verified: `go build`, `go vet`, `staticcheck`, `golangci-lint`, `go test -race ./...` all green.

## Summary by CodeRabbit

* **Improvements**
  * Sent-histogram disk retention extended to 24 hours to reduce premature cleanup.
  * Configuration reloads hardened to avoid shutdown races and event-burst hazards.
  * Filesystem watcher cleanup made more reliable, with improved error handling and lock-release behavior.

* **Tests**
  * Added and updated tests covering retention thresholds, watcher cleanup error handling, and configuration-updater termination scenarios.

### DIFF
--- a/pkg/runner/runner_extra_test.go
+++ b/pkg/runner/runner_extra_test.go
@@ -2359,7 +2359,7 @@ func TestMonitorAndDiskCleaner(t *testing.T) {
 		if err := os.WriteFile(oldFile, []byte("x"), 0o600); err != nil {
 			t.Fatal(err)
 		}
-		oldTime := time.Now().Add(-13 * time.Hour)
+		oldTime := time.Now().Add(-25 * time.Hour)
 		if err := os.Chtimes(oldFile, oldTime, oldTime); err != nil {
 			t.Fatal(err)
 		}
--- a/pkg/runner/runner_extra_test.go
+++ b/pkg/runner/runner_extra_test.go
@@ -2359,7 +2359,8 @@ func TestMonitorAndDiskCleaner(t *testing.T) {
 		if err := os.WriteFile(oldFile, []byte("x"), 0o600); err != nil {
 			t.Fatal(err)
 		}
-		oldTime := time.Now().Add(-25 * time.Hour)
+		// One hour past the retention window, so diskCleaner must remove it.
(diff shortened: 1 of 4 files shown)


linkdata:
@coderabbitai review

pkg/runner/runner_extra_test.go
```diff
@@ -2359,7 +2359,7 @@ func TestMonitorAndDiskCleaner(t *testing.T) {
 		if err := os.WriteFile(oldFile, []byte("x"), 0o600); err != nil {
 			t.Fatal(err)
 		}
-		oldTime := time.Now().Add(-13 * time.Hour)
+		oldTime := time.Now().Add(-25 * time.Hour)
```
linkdata: Should this not be relative to sentHistogramRetention?

linkdata (commented):
Nit: sentHistogramRetention should probably be used for reference

pkg/runner/runner_extra_test.go
linkdata: Good catch — addressed in b2532a2: the offset now reads `-sentHistogramRetention - time.Hour` (one hour past the retention window) instead of the magic `-25 * time.Hour`. Also made `TestDiskCleanerRetentionThreshold` reference the constant the same way (`-sentHistogramRetention ± time.Hour` and `-sentHistogramRetention` at the exact boundary), so the tests stay correct if the retention window ever changes.

linkdata (commented):
LGTM

morkrost (approved):

PR #235, opened 2026-06-12
refactor(runner): migrate lestrrat-go/jwx from v2 to v3

linkdata:
## What

Moves edm's own JWK/JWS code from `lestrrat-go/jwx/v2` to `jwx/v3`. Part of a dependency-reduction series (3 of 5).

## Why

edm currently carries **both** jwx major versions: v3 arrives transitively through `yaronf/httpsign` while edm's own code (EdDSA key loading, MQTT JWS signing, HTTP-signer key export) uses v2. After this PR a single jwx API is used throughout the codebase and v2 is no longer superseded-but-directly-depended-on.

## API changes applied (~40 call sites, mechanical)

| v2 | v3 |
|---|---|
| `jwa.Ed25519`, `jwa.Ed448`, `jwa.EdDSA`, `jwa.RS256` constants | constructor functions `jwa.Ed25519()` etc. |
| `key.Crv()`, `key.Algorithm()`, `key.KeyID()` single return | `(value, bool)` returns |
| `jwk.FromRaw(raw)` | `jwk.Import(raw)` |
| `key.Raw(&dst)` | `jwk.Export(key, &dst)` |
| `jwk.ParseKey`, `key.Set`, `jwk.AlgorithmKey/KeyIDKey`, `jws.Sign(..., jws.WithJSON(), jws.WithKey(...))`, OKP type switches | unchanged |

Two spots that are more than find-and-replace, with behavior preserved:

- `LoadEdDSAJWK` (`pkg/runner/dependencies.go`): `crv, _ = key.Crv()` discards the ok value deliberately — a missing crv yields the zero `EllipticCurveAlgorithm`, which fails the curve check exactly like v2's empty string (commented in code). X25519/X448 keys still fail with `errNotEdDSAJWK`.
- `mqttSignWorker` (`pkg/runner/mqtt.go`): the key's algorithm is read per message; a key with no algorithm set is logged and skipped, matching v2's log-and-continue behavior when `jws.Sign` failed on a missing algorithm. The existing sign-worker tests (including the one mutating the key's alg mid-run) pass unchanged.

## ⚠️ Honest caveat: jwx/v2 stays in go.mod as *indirect*

`httpsign v0.5.1` still imports `jwx/v2` in its main package (it exposes both `NewJWSSigner` and `NewJWSSignerV3`), so v2 moves to the `// indirect` block rather than disappearing, and the module-graph count is unchanged for now. None of edm's compiled call paths use v2 — the only httpsign API edm calls is `NewEd25519Signer`, which is jwx-free. v2 and its v2-only transitives vanish automatically via `go mod tidy` once httpsign drops its v2 code path; filing that upstream issue is a follow-up.

## Operational migration notes

None. Key files, signature formats (JWS for MQTT, RFC 9421 for aggregate-receiver) and on-the-wire behavior are unchanged.

## Verification

- `go vet`, `staticcheck`, `gofumpt -l`, `go build ./...` clean
- `go test -race ./...` green: EdDSA happy path + bad/EC/X25519 negative key cases, MQTT sign/verify round-trips, aggregate-sender tests including the RSA-key rejection path

## Release Notes

* **Improvements**
  * Enhanced message signing robustness with improved handling of incomplete key configurations.
  * Refined error handling in cryptographic operations for better system stability.

* **Updates**
  * Updated cryptographic library dependencies to latest stable version for improved security and compatibility.


linkdata:
@coderabbitai review

linkdata:
@coderabbitai review

linkdata:
@coderabbitai regarding the nitpick on `pkg/runner/aggregate_sender.go` (validate `keyID` before `SetKeyID`): declining for this PR.

The empty-`keyid` behavior is pre-existing — jwx v2's `KeyID()` also returned `""` for a JWK without a `kid`, and this migration's `keyID, _ := signingJwk.KeyID()` deliberately preserves that, so it is not introduced here. Requiring a non-empty `kid` is a behavior change outside the scope of this jwx v2→v3 migration.

The correct place for that check is `LoadEdDSAJWK`, so it covers both the HTTP and MQTT signing keys (both derive identifiers from the key id), rather than only `newAggregateSender`. Tracking it as a separate hardening change.

linkdata:
@coderabbitai update to my earlier reply: rather than tracking this separately, I've folded the hardening into this PR as `83fdb4a`. `LoadEdDSAJWK` now rejects a signing JWK with no key ID (`errJWKMissingKeyID`), which covers both the HTTP and MQTT signing keys since both load through it. So the empty-`keyid` concern is addressed here, not deferred.

linkdata (commented):
LGTM

PR #237, opened 2026-06-12
refactor(cmd)!: replace cobra and viper with stdlib flag and go-toml

linkdata:
## What

Rewrites the CLI on stdlib `flag.FlagSet` with dispatch on `os.Args`, and replaces the Viper configuration pipeline with an explicit `runner.FileConfigProvider`. Part of a dependency-reduction series (5 of 5) — this is the largest module-count win.

> **Stacked PR:** this branch contains the commits of #234 (Config.Validate) and #236 (SIGHUP reload) because it builds directly on both. Review only the last commit, or wait — it will be rebased onto main once #234 and #236 land, after which the diff shows only its own changes.

## Configuration model

`FileConfigProvider.GetConfig()` layers the config explicitly:

```
DefaultConfig()                 ← single source of truth for defaults
  ⊕ strict TOML decode          ← only keys present in the file overwrite;
                                   unknown keys rejected (parity with viper.UnmarshalExact)
  ⊕ immutable override overlay  ← flags set on the CLI or via DNSTAPIR_EDM_* env,
                                   captured once at startup
  → conf.ConfigFile = path; Validate()
```

The overlay is re-applied on every SIGHUP re-read, so **flag > env > file > default precedence holds across reloads by construction** — no re-parsing, no global state.

Mechanics worth knowing for review:

- `newRunFlagSet` binds all 38 run flags directly to a `Config` initialized from `runner.DefaultConfig()`, so flag defaults and the provider's base layer cannot diverge. `mqtt-keepalive` (uint16) uses `fs.Func` + `ParseUint` since stdlib flag has no `Uint16Var`. `mqtt-sign-workers` stays file-only, as today.
- Env handling: `DNSTAPIR_EDM_<FLAG>` values are applied via `fs.Set` **before** `fs.Parse`, so CLI wins; `fs.Visit` afterwards enumerates exactly the env∪CLI-set flags, which become the override overlay via an `overrideFor` switch. `TestOverrideForCoversAllFlags` walks every registered flag to keep that switch from drifting.
- `--config-file` is accepted both **before** the subcommand (`dnstapir-edm --config-file /etc/... run`, the rpm systemd unit's form — stdlib flag parsing stops at the first non-flag argument) and as a run flag, which takes precedence.
- The `Config` struct tags change from `mapstructure:` to `toml:` with identical kebab-case key names — existing config files parse unchanged, including the shipped `dnstapir-edm.toml.sample`.
- Deleted: `ViperConfigProvider`, all ten viper function-pointer test seams, the cobra command tree. `pkg/cmd` test seams are now just `exitProcess`, `userHomeDir`, `osArgs`.

## Dependency impact

Removed from go.mod: `spf13/cobra`, `spf13/viper`, `spf13/pflag`, `spf13/cast`, `spf13/afero`, `sagikazarmark/locafero`, `subosito/gotenv`, `inconshreveable/mousetrap`, `go-viper/mapstructure/v2`, `go.yaml.in/yaml/v3`, and (now that nothing needs it) `fsnotify`. Added as direct: `pelletier/go-toml/v2` (previously indirect via viper) — the only remaining config dependency.

Module graph across the whole series: **155 → 137**.

## ⚠️ Operational migration notes

1. **Invalid values in `DNSTAPIR_EDM_*` environment variables now fail startup** with an error naming the variable, instead of being silently coerced by viper. Audit deployment env for stale/garbage values before rolling out.
2. **`DNSTAPIR_EDM_CONFIG_FILE` now actually selects the config file** (viper never bound the persistent flag, so it was previously ignored). If that variable is set anywhere today, it starts taking effect.
3. The environment is **snapshotted at startup**: SIGHUP reloads keep the startup env values (previously viper re-read env on each access). Changing an env var now requires a restart, which was already the only sane expectation under systemd.
4. Shell completion and cobra's "did you mean" suggestions are gone. `dnstapir-edm help`, `--help` and `run --help` remain.
5. Flag-constraint errors (mutually-exclusive inputs, TLS required-together) now surface at config validation rather than flag parsing — and therefore also catch invalid combinations arriving via the config file or env, which cobra never saw. The TLS rule additionally rejects cert/key files given *without* `input-tls` (cobra enforced this direction; the old validate tags didn't).
6. A `config-file` key *inside* the TOML file is still accepted but no longer load-bearing; `Config.ConfigFile` always reports the file actually read.

## Verification

- `go vet`, `staticcheck`, `gofumpt -l` clean; `go test -race ./...` green
- New tests: dispatch table (help/unknown command/exit codes), config path resolution, full precedence table (defaults / file / env / CLI / unprefixed-env-ignored / uint16 keepalive), overrides surviving a config-file rewrite + re-read, invalid env value error, strict-decode rejection (`toml.StrictMissingError`), missing-file and validation-failure wrapping
- Manual: built binary; bare invocation exit 0 + usage, unknown command exit 1, `run --help`; rpm invocation form `dnstapir-edm --config-file <tmp> run` started the service; SIGHUP reload applied a Crypto-PAn key change with CLI/env overrides intact; SIGTERM clean exit

## Summary by CodeRabbit

* **Refactor**
  * Streamlined command-line interface using standard library tooling
  * Refactored configuration system for improved file-based management and environment variable overrides (with `DNSTAPIR_EDM_*` prefix)
  * Enhanced config precedence: explicit `--config-file`, environment variables, or default `$HOME/.dnstapir-edm.toml`

* **Chores**
  * Updated dependency requirements for lighter footprint

* **Tests**
  * Expanded test coverage for new configuration handling, CLI dispatching, and override precedence logic


linkdata:
@coderabbitai review

linkdata (commented):
LGTM

PR #247, opened 2026-07-01
TMP: Include reloader service that sends SIGHUP after cert renewal

### LINKED ISSUE #245: Make EDM reload configuration from disk
## Background and Motivation

Based on earlier communication, (nowadays) the `dnstapir-edm.service` reloads (relevant) configuration from disk upon a received `SIGHUP`. The `dnstapir-renew.service` does not seam to send such a signal to `dnstapir-edm.service` upon its completion. This needs to change.

## Testing

Manually executing `systemctl kill --signal HUP dnstapir-edm.service` seams to do the trick based on the log-messages generated by `dnstapir-edm.service`. Now the question is how to implement this?

One way is to implement it in `go` inside `cli` which is run by `dnstapir-renew.service`. However we need to make sure that we have permission to send that signal to `dnstapir-edm.service`. Another issue is that we need to have a way to figure out where to send the signal, ie. the `pid` of `dnstapir-edm.service`.

Another way is to implement it using `systemd`, and the only reasonable way I have found as of yet is to create a service `dnstapir-reloader.service` as follows;

    [Unit]
    Description=DNS TAPIR EDM Reloader
    After=dnstapir-renew.service
    
    [Service]
    Type=oneshot
    ExecStart=systemctl kill --signal HUP dnstapir-edm.service
    
    [Install]
    RequiredBy=dnstapir-renew.service

Note that this is running as root, since `systemctl kill` wouldn't let me send the signal like this otherwise. The question is if there is a better way to do this in `systemd` that I've missed? While this adds yet another unit(can we solve the permission issues and integrate this into `dnstapir-renew.service`?) it is kind of smooth using the service name as the reference and let `systemd` handle the signalling.

zluudg:
Attempt to fix #245 

## Summary by CodeRabbit

* **Chores**
  * Added an additional systemd service file to the package so it is installed alongside the existing unit files.

### DIFF
--- a/rpm/dnstapir-edm.spec.in
+++ b/rpm/dnstapir-edm.spec.in
@@ -22,6 +22,7 @@ Source2:       well-known-domains.dawg
 Source3:       ignored.dawg
 Source4:       ignored-ips
 Source5:       dnstapir-edm.sysusers.conf
+Source6:       dnstapir-reloader.service
 BuildRequires: git
 %if 0%{?suse_version}
 BuildRequires: go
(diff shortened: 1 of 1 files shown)


byKryptogram:
I'm missing the `dnstapir-reloader.service` file. Did you miss to include it in the commit @zluudg ?

zluudg:
Work on this will continue in a different branch and a different PR.

PR #256, opened 2026-07-03
runner: make sure that User-Agnet is available on rpm builds

byKryptogram:
When building the rpm-package, the git repo isn't available, hence the version in `debug.ReadBuildInfo()` isn't set properly. Move `main.version` to a separate path `pkg/buildinfo` and use that for the User-Agent.

Fixes: fe8bfca


oej (approved):
Great!

PR #257, opened 2026-07-03
rpm: manage replacement of some changed files in rpm

byKryptogram:
* do not replace the well-known-domains.dawg, ignored.dawg, and ignored-ips files
* do replace systemd service if changed, but save the modified version

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upgrade handling for the service and bundled data files so locally modified settings are less likely to be overwritten during RPM updates.
  * Preserved existing configuration files more safely by treating key files as upgrade-friendly configuration items.


oej (approved):
good work

PR #262, opened 2026-07-13
Work to improve the minimizer

byKryptogram:
**NOT FOR MERGING!**

The goal is to improve the maintainability and performance of the minimiser one step at a time.

Upgrading to newer version of go to enable some new go apis.

Some parts may have impact on the statistics, see commit messages

Also one commit adds comments of potential bugs from before...

Builds on #243 

## Summary by CodeRabbit

* **Improvements**
  * Updated the application’s build environment to Go 1.26.
  * Improved DNS query parsing and handling across supported workflows.
  * Enhanced normalization and matching of domain names, including suffix-based lookups.
  * Improved reliability of background processing, reloads, data collection, and message delivery.
  * Strengthened handling of DNS metadata, network values, and malformed or out-of-range data.


pkg/protocols/protocols.go
```diff
@@ -3,7 +3,7 @@ package protocols
 import (
 	"time"
 
-	"github.com/miekg/dns"
+	"codeberg.org/miekg/dns"
```
zluudg, line 6: Why this switch? I'm not sure v2 of Miekg's lib is considered stable yet.

pkg/runner/wkd.go
```diff
@@ -3,12 +3,11 @@ package runner
...
 	"fmt"
 	"net/netip"
-	"strings"
-	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/miekg/dns"
+	"codeberg.org/miekg/dns"
+	dnsv1 "github.com/miekg/dns"
```
zluudg, line 11: Using two versions of the same lib? Do we really want to go there?

go.mod
```diff
@@ -1,8 +1,9 @@
 module github.com/dnstapir/edm
 
-go 1.25.6
+go 1.26
 
 require (
+	codeberg.org/miekg/dns v0.6.83
```
zluudg: I do not think we are ready for this migration of a core component. A strong motivation is needed.

zluudg (commented):
So far I've identified three seemingly unrelated changes in this PR: migration of miekgs lib to a new version, change of golang version and formatting error strings. Maybe some things can be made into their own PRs to make reviewing easier. Like changing miekgs lib and the language version.

pkg/runner/wkd.go
byKryptogram: This is a temporary measure, some of the helper functions have changed slightly and I've not verified that they don't impose a problem for us. Therefore I've left then on the v1 version for the moment. I'm trying to do an incremental switch.

byKryptogram:
> So far I've identified three seemingly unrelated changes in this PR: migration of miekgs lib to a new version, change of golang version and formatting error strings. Maybe some things can be made into their own PRs to make reviewing easier. Like changing miekgs lib and the language version.

Sorry for the confusion! I'm using this *draft* PR to help myself keep track of things and trigger the GitHub actions. Not sure if there are better ways to do it.

go.mod
byKryptogram: We have an issue of the EDM being slow, or in other words; using a lot of resources. Switching to the new version of miekg's dns library does help with that. Moreover the new version lets us do partial parsing of the dns message if we choose to. Today we don't use the Answers/Authorities/Additionals sections in the analysis.

Some benchmark of this [example](https://pkg.go.dev/golang.org/x/net/dns/dnsmessage#example-Parser) `msg`, the three first are comparable parsing the message fully. If we skip everything after the Questions section we can improve it even further(the last two lines). The third library is [golang.org/x/net/dns/dnsmessage](https://pkg.go.dev/golang.org/x/net/dns/dnsmessage)

```
BenchmarkMiekgV1-32                      	  524742	      2108 ns/op	     672 B/op	      22 allocs/op
BenchmarkMiekgV2-32                      	  745074	      1604 ns/op	     520 B/op	      16 allocs/op
BenchmarkGoDnsMessage-32                 	  601540	      1964 ns/op	    2160 B/op	       8 allocs/op

BenchmarkMiekgV2OnlyQuestions-32         	 2418212	       509.7 ns/op	     184 B/op	       6 allocs/op
BenchmarkGoDnsMessageOnlyQuestions-32    	 2112134	       560.2 ns/op	     864 B/op	       2 allocs/op
```

pkg/protocols/protocols.go
byKryptogram: [Reason see elsewhere.](https://github.com/dnstapir/edm/pull/262#discussion_r3571661287) In the [README](https://codeberg.org/miekg/dns#modern-lightweight-dns-library) of the new library it states
* "Complete and usable DNS library."
* "We try to keep the "main" branch as sane as possible and at the bleeding edge of standards, avoiding breaking changes wherever reasonable. But because this version is young, we allow ourselves some more headroom for making backwards incompatible changes."
* "This new version will not soon see a v1.0.0 release because I want to be able to still make changes. In a year or two (2028?) when things have stabilized it will be blessed with a v1.0.0."

While the old version's [README](https://github.com/miekg/dns/blob/master/README.md) states "The version here will only see specific fixes and nothing else. At some point this repo will be archived."

pkg/runner/session_test.go
```diff
@@ -449,8 +448,7 @@ func TestSessionWriterLogsCreateError(t *testing.T) {
 
 func TestNewSessionAllowsMissingSocketMetadata(t *testing.T) {
 	edm := discardEDM()
-	msg := new(dns.Msg)
-	msg.SetQuestion("example.com.", dns.TypeA)
+	msg := dns.NewMsg("example.com.", dns.TypeA)
```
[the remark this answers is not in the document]
byKryptogram: That is incorrect!

pkg/runner/wkd_test.go
```diff
@@ -329,7 +299,7 @@ func TestSendUpdateBranches(t *testing.T) {
 	cases := []struct {
 		name    string
 		ipBytes []byte
-		rcode   int
+		rcode   uint16
```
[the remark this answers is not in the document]
byKryptogram: Also incorrect.

byKryptogram:
test pr not used anymore

PR #266, opened 2026-07-13
Simplify implementation using append

byKryptogram:
Should be no functional difference, just less code see [`append`](https://pkg.go.dev/builtin#append).


zluudg (approved):
LGTM

PR #267, opened 2026-07-13
Help dns.NextLabel by skipping ahead one label at a time

byKryptogram:
Should be no functional difference, but a bit quicker.

Documentation: [`dns.NextLabel`](https://pkg.go.dev/github.com/miekg/dns#NextLabel)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS label suffix matching to ensure the correct index is used when checking domain names.
  * No changes to public APIs or user-facing functionality.

### DIFF
--- a/pkg/runner/wkd.go
+++ b/pkg/runner/wkd.go
@@ -77,11 +77,13 @@ func getDawgIndex(dawgFinder dawg.Finder, name string) (int, bool) {
 		// Next try to look up suffix matches, so for the name
 		// "www.example.com." we will check for the strings
 		// ".example.com." and ".com.".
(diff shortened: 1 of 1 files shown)


zluudg (commented):
Why is not using the interface of dns.NextLabel as intended good enough for us?

If we are going to do things in a convoluted way here I think a reason should be given. Preferably as a code comment.

byKryptogram:
Initially I got a clear, although low wall time, improvement on `x86_64`. While doing more testing on `x86_64` and `arm64` the later thanks to @eest we have now arrived at a point where it is hard to say which version is best without digging into the generated assembly. This is not worth the effort for a potential improvement this small.

